PR #4, opened 2025-09-15
Proposal: browser lifecycle events doc

joaquin-diaz:
# Proposal: Browser Lifecycle Events

Adds a conceptual model for **browser lifecycle events**, organized into **Load**, **User Interaction**, and **Unload** phases. Includes an event list with **purpose**, **frequency**, and **status** for both semantic conventions and instrumentation.

## Goal

Standardize browser telemetry interpretation and guide instrumentation authors.


docs/browser-lifecycle.md
```diff
@@ -0,0 +1,74 @@
...
+| `browser.error` | Captures unhandled JavaScript errors. | Not created | Not created                                                                                                                                                   |
+| `browser.console` | Captures browser console messages such as warnings and logs. | Not created | Not created                                                                                                                                                   |
+| `browser.user_interaction` | Captures user input events (clicks, scrolls, keypresses) and their latency. | Not created | Not created                                                                                                                                                   |
+| `browser.performance_resource_timing` | Captures late-loaded assets (ads, analytics, lazy images, etc.). | Not created | Merged (similar) [instrumentation-document-load](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-document-load) |
```
martinkuba, line 62: @Karlie-777 has an existing PR for semantic conventions [here](https://github.com/open-telemetry/semantic-conventions/pull/1943).

docs/browser-lifecycle.md
```diff
@@ -0,0 +1,74 @@
...
+| `browser.web_vital` | Captures Web Vitals metrics such as CLS, INP, and FID that occur after page load. | Merged [WebVitals](https://opentelemetry.io/docs/specs/semconv/browser/events/#webvital-event) | Not created                                                                                                                                                   |
+| `browser.error` | Captures unhandled JavaScript errors. | Not created | Not created                                                                                                                                                   |
+| `browser.console` | Captures browser console messages such as warnings and logs. | Not created | Not created                                                                                                                                                   |
+| `browser.user_interaction` | Captures user input events (clicks, scrolls, keypresses) and their latency. | Not created | Not created                                                                                                                                                   |
```
martinkuba, line 61: There is an existing PR [here](https://github.com/open-telemetry/semantic-conventions/pull/1941) (also from @Karlie-777).

docs/browser-lifecycle.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+### Unload Phase
+
+**Starts when the user initiates a navigation away, closes the tab, or reloads**  
+Captures the start of the unload process.
+
+| Event | Description | Semantic Conventions Status | Instrumentation Status |
+|---|---|---|---|
+| `browser.page_unload` | Signals the start of the page unload process. Occurs once per page unload. | Not created | Not created |
```
martinkuba, line 73: I am not sure if this event is necessary. What value would it provide?

docs/browser-lifecycle.md
```diff
@@ -0,0 +1,74 @@
...
+| `browser.page_view` | Signals the start of a **hard** page navigation. Occurs once per page load. | In review [PR1910](https://github.com/open-telemetry/semantic-conventions/pull/1910) | In review [PR2386](https://github.com/open-telemetry/opentelemetry-js-contrib/pull/2386)                                                             |
+| `browser.navigation_timing` | Captures detailed technical milestones from the [PerformanceNavigationTiming](https://developer.mozilla.org/docs/Web/API/PerformanceNavigationTiming) API. Occurs once per page load. | In review [PR1919](https://github.com/open-telemetry/semantic-conventions/pull/1919) | Not created                                                                                                                                          |
+| `browser.page_load` | Summarizes the user-visible page load experience from navigation start until LCP (includes total duration, paint timings, resource counts, transfer size). Occurs once per page load. | Not created | Not created                                                                                                                                          |
```
martinkuba, line 43: Is the main intent of this event to summarize all the other events that happen during page load to make it easier for backends?

Alternatively
- duration is part of navigation timing (if based on the window load event)
- FCP and LCP are part of web vitals
- resource counts could be determined from resource timing data
- transfer size is available in resource timing data

docs/browser-lifecycle.md
```diff
@@ -0,0 +1,74 @@
...
+2. **User interaction phase** – from LCP until the page begins unloading
+3. **Unload phase** – when the page begins unloading
+
+---
+
+## Lifecycle Phases
+
+### Load Phase
+
+**Starts at navigation start → Ends at Largest Contentful Paint (LCP)**
```
BenoitZugmeyer, line 36: **suggestion:** I would avoid using the LCP as a structuring part of the model:

* it's not supported in all browsers (ex: Safari)
* It implies that we actually wait for a user interaction before going to the next phase, since LCP entries are notified until a user interaction happens (the Core Web Vital LCP is the time until the latest LCP). This could be problematic on pages that don't have user interactions (ex: a page opened in background).
* LCP can happen before the page actually finishes to load (before the `load` event), so correlating to the "Load" phase could be confusing.

docs/browser-lifecycle.md
martinkuba: The unload event is basically deprecated and should not be used due to bfcache. The pagehide event fires also when a document is put into bfcache, which means that the same document can be restored. This implies that it can have a more complex lifecycle than just unload.

martinkuba: There is also the visibilitychanged event. This does not indicate the end of a document lifecycle (page view?), but could be useful to understand when the user is actually on the page.

joaquin-diaz:
Closing, following in https://github.com/open-telemetry/opentelemetry-browser/pull/6 from fork

PR #35, opened 2025-10-24
feat: add UserActionInstrumentation

joaquin-diaz:
# Goal

Add the first iteration of the user action instrumentation only covering the *mousedown* event based on the current [semantic conventions PR](https://github.com/open-telemetry/semantic-conventions/pull/2992) 

### Vitest 

Simple vitest setup for running unit tests using jsdom

### Initial monorepo setup

Preparing for a future turbopack integration, I set up some packages for the upcoming monorepo setup. 
* **instrumentation-user-action**: package where the new instrumentation lives
* **test-utils**: a private shared package where test utilities can live
* **web-utils**: a private shared package where cross instrumentation web utils can live

Each package has these set of scripts:
* **test**: vitest run
* **test:watch**: vitest
* **test:coverage**: vitest --coverage

Once we set up a monorepo tool (possible turbopack) we can easily run these commands from root 

TODO: make sure package can build

## What's **not** included

* CI/CD: still not hooked up, will wait until we have turbopack 
* Release/publishing: not yet; we need to review the release strategy with the sig

### DIFF
--- a/packages/instrumentation-user-action/src/instrumentation.ts
+++ b/packages/instrumentation-user-action/src/instrumentation.ts
@@ -95,7 +95,7 @@ export class UserActionInstrumentation extends InstrumentationBase<UserActionIns
         [ATTR_TAG_NAME]: element.tagName,
         [ATTR_TAGS]: otelPrefixedAttributes,
         [ATTR_MOUSE_EVENT_BUTTON]: this._getMouseButtonFromMouseEvent(event),
-        [ATTR_CSS_SELECTOR]: xPath,
+        [ATTR_CSS_SELECTOR]: cssSelector,
       },
     });
   };


packages/web-utils/src/getElementXPath.ts
```diff
@@ -0,0 +1,78 @@
...
+    const sameName = (n as HTMLElement).localName === element.localName;
+    return sameType && sameName;
+  });
+
+  return siblings.length > 1 ? siblings.indexOf(element) + 1 : 0;
+};
+
+const getNodeValue = (element: HTMLElement, optimized = false): string => {
+  const type = element.nodeType;
+  if (type === Node.ELEMENT_NODE) {
```
BenoitZugmeyer: Nit: Since element is an `HTMLElement`, checking its type seems superfluous.

packages/instrumentation-user-action/src/instrumentation.ts
```diff
@@ -0,0 +1,112 @@
...
+    const otelPrefixedAttributes: Record<string, string> = {};
+
+    // Grab all the attributes in the element that start with data-otel-*
+    for (const attr of element.attributes) {
+      if (attr.name.startsWith(OTEL_ELEMENT_ATTRIBUTE_PREFIX)) {
+        otelPrefixedAttributes[
+          attr.name.slice(OTEL_ELEMENT_ATTRIBUTE_PREFIX.length)
+        ] = attr.value;
+      }
+    }
```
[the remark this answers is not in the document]
BenoitZugmeyer: As long as it is consistent with the semconv, I'm fine with it. Although in our SDK (Datadog), when we look at attributes similar to this, we usually:

* do iterate on all ancestors
* don't include siblings or children

I think it makes sense, for example imagine a case like:

```html
<button data-otel-checkout="true">
  <img src="./icon.svg" />
  Checkout
</button>
```

Users would likely expect to have the `checkout: "true"` attribute even if the click target is the `img`.

packages/web-utils/package.json
martinkuba: There is an existing [web-common package](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/web-common). It might be confusing to have two similar packages published in npm. I suggest that we either move web-common to this repo, or add these utils there for now.

PR #72, opened 2025-12-04
chore: add turbo generator for instrumentation

overbalance:
> [!NOTE]
> Licensing header decisions from #88 will impact the content of this PR.

## Summary
- Add turbo generator for scaffolding new instrumentation packages
- Includes templates for package.json, tsconfig, tsdown config, source files, and test scaffold
- Generated packages follow existing patterns from instrumentation-user-action
- Uses short SPDX license header format from #88
- Add skipLibCheck to fix type conflicts from @turbo/gen dependencies
- Update turbo to 2.6.3

## Test plan
- Run `npm run generate` to scaffold a new package
- Verify generated files match expected structure


turbo/generators/templates/instrumentation/index.ts.hbs
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
david-luna: could we use the short form like we are doing in https://github.com/open-telemetry/opentelemetry-browser/pull/88?

turbo/generators/config.ts
```diff
@@ -0,0 +1,64 @@
+import type { PlopTypes } from '@turbo/gen';
+
+export default function generator(plop: PlopTypes.NodePlopAPI): void {
+  // Instrumentation Package Generator
+  plop.setGenerator('instrumentation', {
+    description: 'Create a new browser instrumentation package',
+    prompts: [
+      {
+        type: 'input',
+        name: 'name',
```
david-luna: is there a way to check that the name is not taken? Would it fail if I use an existing name or would it overwrite the files?

turbo/generators/templates/instrumentation/index.ts.hbs
overbalance: I prefer the short form. Still waiting on approvals.

david-luna (approved):

PR #98, opened 2025-12-17
feat: add `@opentelemetry/instrumentation-console` package for capturing console calls as OpenTelemetry logs

### LINKED ISSUE #15: Instrumentation to capture console logs as OTel logs
### Is your feature request related to a problem? Please describe
This is a feature request to capture console logs and send them to OTLP backends using the OTel Logs API/SDK.

### Describe the solution you'd like to see
User should be able to capture console logs and export them to OpenTelemetry backends.

Related to https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1558 and https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1559

drdreo:

Users should be able to capture console logs and export them to OpenTelemetry. This PR adds a browser console method instrumentation which sends them as opentelemetry logs

Fixes #15 

## Type of change

## How Has This Been Tested?

Unit tests for console logs being sent to the in memory processor as well as the config options were verified in unit tests with edge case testing for the custom serializer failing.

- [x] Unit tests

## Checklist:

## OPEN QUESTIONS
what semantic attributes?
 - browser.console for event  and browser.console.method for type

severity numberfor log vs info?
-  INFO = 9?

is the traceparent of the current active span enough? 
- from my testing that can get disconnected when the frontend was served by the backend and we could use the traceparent from meta tags


drdreo:
@overbalance happy new year. I got back at this PR and fixed the linting error. May you reapprove the workflow please?

overbalance:
@drdreo happy new year! the license header rules changed on main. to repair please remove the licenses at the top of your files and `npm run lint` to add them back automatically. I'll run the instrumentation and continue my review.

overbalance:
and `npm i` to update the lockfile @drdreo

packages/instrumentation-console/src/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import type { SpanContext } from '@opentelemetry/api';
+import { parseTraceParent, TRACE_PARENT_HEADER } from '@opentelemetry/core';
+
+export function getTraceParent(): SpanContext | null {
+  const metaElement = Array.from(document.getElementsByTagName('meta')).find(
+    (e) => e.getAttribute('name') === TRACE_PARENT_HEADER,
+  );
+  return parseTraceParent(metaElement?.content || '');
+}
```
overbalance: You can access the meta tag without a loop:
```suggestion
export function getTraceParent(): SpanContext | null {
  const metaElement = document.querySelector<HTMLMetaElement>(
    `meta[name="${TRACE_PARENT_HEADER}"]`
  );
  return parseTraceParent(metaElement?.content ?? '');
}
```

overbalance (changes requested):
Code looks good. Let me know if you have concerns about my suggestion.

One issue I ran into was an infinite loop if I use the recommended OTel example https://opentelemetry.io/docs/languages/js/getting-started/browser/#creating-an-exporter

Would you mind adding a section to the docs stating the conflict if you also use `ConsoleSpanExporter` during development?

packages/instrumentation-console/src/instrumentation.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        instrumentation.logger.emit({
+          body,
+          eventName: CONSOLE_LOG_EVENT_NAME,
+          severityNumber: SEVERITY_MAP[method],
+          severityText: method,
+          attributes: {
+            [ATTR_CONSOLE_METHOD]: method,
+            ...(spanContext && {
+              'trace.id': spanContext.traceId,
```
martinkuba: You can just pass in the [context](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/api-logs/src/types/LogRecord.ts#L90) itself, and [traceId](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L199) and [spanId](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L213) get exported as top-level fields rather than attributes.

drdreo:
Hej, i've addressed your feedback. I added a simple vite test app at `instrumentation-console/examples` to verify the instrumentation - let me know if you like to have some example app or if i should remove it again.

packages/instrumentation-console/examples/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview"
+  },
+  "devDependencies": {
+    "@opentelemetry/sdk-trace-base": "2.2.0",
+    "@opentelemetry/sdk-trace-web": "2.2.0",
+    "typescript": "~5.9.3",
+    "vite": "^7.2.6"
```
overbalance:
```suggestion
    "build": "vite build",
    "preview": "vite preview"
```

I will add this folder to workspaces and update the example after merge.

overbalance (commented):
Re-request me for review once the conversation with Martin resolves.

packages/instrumentation-console/src/instrumentation.test.ts
```diff
@@ -0,0 +1,258 @@
...
+import { afterEach, beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { ConsoleInstrumentation } from './instrumentation.ts';
+import { ATTR_CONSOLE_METHOD, CONSOLE_LOG_EVENT_NAME } from './semconv.ts';
+
+describe('ConsoleInstrumentation', () => {
+  let inMemoryExporter: InMemoryLogRecordExporter;
+  let instrumentation: ConsoleInstrumentation;
+
+  beforeAll(() => {
+    inMemoryExporter = setupTestLogExporter();
```
martinkuba: Running the tests is leaking all of the console messages from tests. This could be resolved by mocking the console object, something like this here:

```js
    globalThis.console = {
      error: () => {},
      log: () => {},
      info: () => {},
      warn: () => {},
      trace: () => {},
      debug: () => {},
    } as unknown as Console;
```

martinkuba (changes requested):

packages/instrumentation-console/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "check-types": "tsc",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "test:coverage": "vitest --coverage"
+  },
+  "peerDependencies": {
+    "@opentelemetry/api": "^1.0.0"
+  },
+  "dependencies": {
+    "@opentelemetry/api-logs": "0.208.0",
```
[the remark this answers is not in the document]
overbalance: I initially set the versions to star to make sure that the local version was linked between packages. When publishing we will need to replace the star with actual versions though. I know Changesets has that feature. Should we copy the release process from core repo or should we try a newer tool like Changesets?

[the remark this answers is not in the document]
martinkuba: Gotcha, we can keep the same process for now.

packages/instrumentation-console/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@opentelemetry/instrumentation": "0.208.0",
+    "@opentelemetry/web-utils": "*"
```
martinkuba: I think this dependency can now be removed?

packages/instrumentation-console/README.md
```diff
@@ -0,0 +1,115 @@
...
+    new ConsoleInstrumentation(),
+  ],
+});
+
+// Now all console calls will be captured as OpenTelemetry logs
+console.log('Hello, World!');
+console.error('Something went wrong!');
+```
+
+> **Warning:** Do not use this instrumentation with the `ConsoleSpanExporter`. This creates an infinite loop where console logs from the exporter itself are captured as new logs, creating more console output. See [OpenTelemetry Browser Exporters](https://opentelemetry.io/docs/languages/js/getting-started/browser/#creating-an-exporter) for available options.
```
martinkuba: The same applies to `ConsoleLogRecordExporter`, and the example above uses ConsoleLogRecordExporter - can you please update the example to use the [OTLPLogExporter](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/exporter-logs-otlp-http)?

martinkuba (approved):

overbalance (commented):
@drdreo There was an issue in another instrumentation recently where repeatedly calling enable() would add listeners over and over. I think this code may be subject to that as well. Does this also need a guard?

packages/instrumentation-console/README.md
```diff
@@ -0,0 +1,114 @@
...
+    new ConsoleInstrumentation(),
+  ],
+});
+
+// Now all console calls will be captured as OpenTelemetry logs
+console.log('Hello, World!');
+console.error('Something went wrong!');
+```
+
+> **Warning:** Do not use this instrumentation with the `ConsoleSpanExporter` or `ConsoleLogRecordExporter`. This creates an infinite loop where console logs from the exporter itself are captured as new logs, creating more console output. Use exporters that send data over the network, such as `OTLPLogExporter`, `OTLPTraceExporter`, or other non-console exporters. See [OpenTelemetry Browser Exporters](https://opentelemetry.io/docs/languages/js/getting-started/browser/#creating-an-exporter) for available options.
```
pichlermarc: A `DiagConsoleLogger` registered with `diag` from `@opentelemetry/api` will likely cause the same issue, even when the OTLP exporters. That's the reason why there was no console instrumentation in core/contrib yet - it essentially means you need to give up any attempt at troubleshooting when using the instrumentation. 

IMO a `console` instrumentation is not viable until that problem is solved.

pichlermarc: > one shouldn't enable a console exporter to debug in production to begin with.

I agree. My comment is about the `diag` logger, not the `ConsoleLogRecordExporter`. In fact, the function that is used by the console exporter `console.dir()` is not instrumented here at all, so the `ConsoleLogRecordExporter` is not the one causing the infinite loop, `diag` logging is.

The `diag` logger is the *internal* logging of the API+SDK, as well as any and that's commonly turned on in production (at least for errors) from what I have seen.

martinkuba: @pichlermarc @dyladan I have created a draft of a `suppressLogging` mechanism [here](https://github.com/open-telemetry/opentelemetry-js/pull/6355).

I added it to the core package to be consistent with `suppressTracing`. The downside is that the default `DiagLogger` is in the API package. I could either move this to the API package and add it directly to the default logger, or users will have to wrap the logger by using the `createSuppressedDiagLogger` I have drafted in the PR.

Forcing users to use `createSuppressedDiagLogger` does not seem that different than adding a different logger in this package, but maybe less likely to miss. 🤷  What do you think?

drdreo: To circle back now, are we good to proceed or is there anything from my side to drive this further?

overbalance (approved):

packages/instrumentation/src/console/instrumentation.ts
```diff
@@ -0,0 +1,110 @@
...
+        return original.apply(this, args);
+      } as Console[ConsoleMethod];
+    };
+  }
+
+  override enable(): void {
+    const methods = this._getLogMethods();
+    for (const method of methods) {
+      if (typeof console[method] === 'function') {
+        this._wrap(console, method, this._patchConsoleMethod(method));
```
david-luna, line 111: nit: Although it's unlikely that another instrumentation patches console methods there was a discussion recently since there is a potential incompatibility between instrumentations when they unpatch APIs.
The conclusion was that instrumentations should patch the API once and the patched functions should check for the `enabled` state of the instrumentation. ref: https://github.com/open-telemetry/opentelemetry-browser/issues/204#issuecomment-4096532121

david-luna (commented):
Overall looks good. https://github.com/open-telemetry/opentelemetry-browser/pull/175 moved all instrumentations in a single package I think this one should live there too.

#### Checklist 
- [x] move the code into `/packages/instrumentation/src/console`
- [x] set the scope name to `@opentelemetrry/browser-instrumentation/console`

david-luna (changes requested):

overbalance (changes requested):
Oops. I missed what folder this was in.

packages/instrumentation/src/console/instrumentation.ts
drdreo: I have moved the console instrumentation to the new folder structure and read through the unpatching discussion. So no unwrap on disable anymore, just is active false

packages/instrumentation-console/README.md
pichlermarc: good to proceed. feel free to resolve the conversation :) 
thank you everyone for your patience :)

packages/instrumentation/src/console/instrumentation.ts
david-luna: @drdreo thanks for doing it :) 

I'd leverage this conversation to point to another topic/question which is if the configuration should be static or could change in time. In this instrumentation I see the console methods option is final and if the user excluded a method there is no way to change that unless the config is changed and the application reloaded.

IMHO if a instrumentation can observe changes in configuration and behave accordingly the user has more control on the telemetry generated and even can implement mechanisms to tune/config instrumentations without the need of a reload or re-deploy. A scenario that comes to mind is an app containing a bug which spits a ton of errors in the console (like in an endless loop). If the app is accessed my many users this error rate (logs sent to OTLP endpoint) could be multiplied by the # of users. In that case I wish I had a way to tell the instrumented apps to stop sending logs, at least of error severity.

david-luna: Worth mention with "dynamic config" we should have only one instance of the instrumentation for all the test and just adjust the configuration for the feature we want to test. I think having duplicated instances of the same instrumentation in the same web page is not desired.

drdreo: Hmm, i think we could easily add that to dynamically check the configured methods on call and just patch all from the start.
Want me to add it here or create a new feature PR?

david-luna: @drdreo 

would you consider doing it for this PR? I'm okay if we do it in a follow up PR

package-lock.json
```diff
@@ -345,9 +345,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
```
david-luna, line 361: which `npm` version did you use while developing this? IIUC it might be related to https://github.com/npm/cli/issues/8514 and fixed in npm@11.11.0 (arborist@9.4.0) https://github.com/npm/cli/pull/9006

I think its better to keep the `package-lock.json` as is.

david-luna (changes requested):

package-lock.json
overbalance: this keeps happening when renovate does "lockfile maintenance" too

overbalance (approved):

david-luna (approved):

package-lock.json
drdreo: I have reverted the lockfile now. i used to be on npm@11.9.0

packages/instrumentation/src/console/instrumentation.ts
drdreo: @david-luna i have added dynamic config for log methods in this PR :)

david-luna:
@drdreo thank you for your contribution :)

drdreo:
> @drdreo thank you for your contribution :)

 @david-luna Thanks for having me :) glad to contribute

overbalance:
@drdreo Thanks again, this has been released. It would be great to add an entry to the readme inside the instrumentation folder.

martinkuba:
> It would be great to add an entry to the readme inside the instrumentation folder.

I have created https://github.com/open-telemetry/opentelemetry-browser/issues/260, which is related

PR #111, opened 2026-01-07
feat: add performance navigation timing instrumentation

### LINKED ISSUE #16: Implement browser Page Navigation Timing event instrumentation
This should be a new instrumentation package. The instrumentation should use only the [Events API](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/api-events?rgh-link-date=2024-04-23T17%3A20%3A29Z) as a dependency for generating the event.

The event semantic conventions are drafted [here](https://github.com/scheler/semantic-conventions/blob/browser-events-v2/docs/browser/events.md#pagenavigationtiming) (and part of https://github.com/open-telemetry/semantic-conventions/pull/838).

codecapitano:

Fixes #16 

Implements tracking of [PerformanceNavigationTiming](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTimings). This enables users to monitor rendering and load performance metrics for page loads (hard navigations).

## Type of change

## How Has This Been Tested?

- [x] Unit tests

## Checklist:


packages/instrumentation-navigation-timing/src/instrumentation.ts
```diff
@@ -0,0 +1,229 @@
...
+    return [];
+  }
+
+  override enable(): void {
+    if (
+      typeof performance === "undefined" ||
+      typeof performance.getEntriesByType !== "function"
+    ) {
+      return;
+    }
```
overbalance: Guard should be removed, performance is Widely Available

packages/instrumentation-navigation-timing/src/instrumentation.test.ts
```diff
@@ -0,0 +1,348 @@
...
+  let instrumentation: NavigationTimingInstrumentation;
+  let restoreReadyState: (() => void) | undefined;
+  let restoreGetEntriesByType: (() => void) | undefined;
+
+  beforeAll(() => {
+    inMemoryExporter = setupTestLogExporter();
+  });
+
+  beforeEach(() => {
+    instrumentation = new NavigationTimingInstrumentation();
```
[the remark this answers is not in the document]
codecapitano: Done

packages/instrumentation-navigation-timing/src/instrumentation.ts
```diff
@@ -0,0 +1,229 @@
...
+    // If the document is still loading, wait for `load` and try again.
+    if (document.readyState !== "complete") {
+      window.addEventListener("load", this._onLoad, { once: true });
+      return;
+    }
+
+    // If the document is already complete but navigation timings are not finalized yet,
+    // do a single deferred re-check to allow the browser to finish populating the entry.
+    if (!this._completeDelayTimeoutId) {
+      this._completeDelayTimeoutId = window.setTimeout(() => {
```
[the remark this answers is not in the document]
codecapitano: Great catch, we should take care of it I can at least immediately think about pages loading background where some might delay the load event after readyState is complete.

.gitignore
```diff
@@ -15,5 +15,8 @@ dist
 # IntelliJ based IDEs
 .idea
 
+# VSCode
+.vscode
```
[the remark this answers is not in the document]
overbalance: Thanks for adding. Lint is designed to run pre-commit so you don't have to customize your IDE just for this codebase. 

I'll see if I can replicate so it doesn't trip anyone else up.

packages/instrumentation-navigation-timing/src/instrumentation.test.ts
```diff
@@ -0,0 +1,428 @@
...
+    expect(instrumentation).toBeInstanceOf(NavigationTimingInstrumentation);
+  });
+
+  it('should enable and disable without errors', () => {
+    expect(() => {
+      instrumentation.enable();
+      instrumentation.disable();
+    }).not.toThrow();
+  });
+
```
overbalance: Here's the case I was referring to:

```suggestion

  it('should not add duplicate listeners when enable() is called twice', () => {
    const addEventListenerSpy = vi.spyOn(window, 'addEventListener');
    setReadyState('loading');

    instrumentation.enable();
    instrumentation.enable(); // Second call should be a no-op

    const pagehideCalls = addEventListenerSpy.mock.calls.filter(
      ([event]) => event === 'pagehide',
    );
    expect(pagehideCalls.length).toBe(1); // Should only be 1, not 2

    addEventListenerSpy.mockRestore();
  });
  
```

overbalance (changes requested):
Looks like some vitest browser screenshots were accidentally checked in also.

packages/instrumentation-navigation-timing/src/instrumentation.test.ts
codecapitano: Done

packages/instrumentation-navigation-timing/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "watch": "tsdown --watch",
+    "check-types": "tsc",
+    "test": "vitest run --browser.enabled=false",
+    "test:watch": "vitest --browser.enabled=false",
+    "test:coverage": "vitest --coverage --browser.enabled=false"
+  },
+  "dependencies": {
+    "@opentelemetry/api-logs": "*",
+    "@opentelemetry/instrumentation": "*",
+    "@opentelemetry/web-utils": "*"
```
overbalance:
```suggestion
    "@opentelemetry/instrumentation": "*"
```

overbalance (approved):

Issue #18, opened 2023-04-05
Core Web Vitals Plugin


austinlparker:
### Is your instrumentation request related to a problem? Please describe
[Core Web Vitals](https://web.dev/vitals/) is a widely accepted and utilized set of measurements for web client performance. Frontend developers rely on CWV in creating SLIs and SLOs for their applications.

### Is it applicable for Node or Browser or both?
Browser

### Do you expect this instrumentation to be commonly used?
I would expect this to be come a default part of the web instrumentation.

### What version of instrumentation are you interested in using?
n/a

### Additional context
I'm mostly adding this request as I don't see it anywhere already and wanted to have a public record/discussion of it. Perhaps this has been discussed in the SIG previously?

david-luna:
@martinkuba I see you did some changes here. I'll try to allocate some time to work on this. We could discuss it in the SIG

PR #145, opened 2026-02-06
feat(instrumentation): add web-vitals instrumentation

overbalance:

Adds Core Web Vitals instrumentation to OpenTelemetry Browser, enabling automatic collection of performance metrics as OpenTelemetry log records.

Fixes https://github.com/open-telemetry/opentelemetry-browser/issues/18

~~Note: the dev server/example on this branch has been extracted to https://github.com/open-telemetry/opentelemetry-browser/pull/171 and will naturally be dropped from this branch after it's merged.~~ Merged

- Add `@opentelemetry/instrumentation-web-vitals` package
- Collect INP, CLS, TTFB, LCP, FCP via `web-vitals` library
- Emit metrics as OpenTelemetry log records
- Optionally include attribution data as JSON in the log body

## Type of change

## How Has This Been Tested?

- [x] Unit tests for all metric collection and optional attribution data

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,8 +31,8 @@
         "publint": "0.3.17",
         "tsdown": "0.20.3",
         "turbo": "2.8.10",
-        "typescript-eslint": "8.56.0",
-        "vite": "^8.0.0-beta.15",
+        "typescript-eslint": "8.56.1",
+        "vite": "8.0.0-beta.16",
         "vitest": "4.0.18"
       },
       "engines": {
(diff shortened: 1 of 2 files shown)


packages/instrumentation-web-vitals/src/index.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export { WebVitalsInstrumentation } from './instrumentation.ts';
+export {
```
david-luna: [question] Are this semconv expected to be used by consumers of this package?

packages/instrumentation-web-vitals/src/semconv.ts
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * This file contains a copy of unstable semantic convention definitions
```
jpmunz: curious as to having the copy here vs. depending on that package directly?

packages/instrumentation-web-vitals/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@opentelemetry/instrumentation-web-vitals",
+  "version": "0.1.0",
```
jpmunz: would it be possible to release this as "1.0.0"? Is there a particular change or milestone we would want to hit prior to doing that?

packages/instrumentation-web-vitals/src/semconv.ts
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_WEB_VITAL_ATTRIBUTION_CONNECTION_DURATION =
+  'browser.web_vital.attribution.connection_duration';
+
+/**
+ * @experimental This attribute is experimental and is subject to breaking changes in minor releases of `@opentelemetry/semantic-conventions`.
+ */
+export const ATTR_WEB_VITAL_ATTRIBUTION_REQUEST_DURATION =
```
vigneshshanmugam: Any reason for having these defined under attribution? Why not clearly define the top level field for each of the web-vitals? Something like `browser.web_vitals.ttfb.total, browser.web_vitals.ttfb.dns_duration`

packages/instrumentation-web-vitals/src/semconv.ts
martinkuba: This is based on the guidance described here
https://github.com/open-telemetry/opentelemetry-js/tree/main/semantic-conventions#unstable-semconv

packages/instrumentation-web-vitals/src/semconv.ts
martinkuba: The [existing semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/model/browser/events.yaml) do not include all of the attribution attributes. I think this is fine, but we should follow up by adding them in the semantic conventions repo.

packages/instrumentation-web-vitals/src/instrumentation.ts
```diff
@@ -0,0 +1,276 @@
...
+    this._isEnabled = false;
+    this._diag.debug('Instrumentation disabled, pausing emission');
+  }
+
+  /**
+   * Gets the timestamp for a metric based on attribution timing.
+   * Returns undefined to let OTel use the current time for metrics without
+   * specific timing information.
+   */
+  private _getTimestampForMetric(
```
martinkuba: I think this should include LCP as well. The value of LCP is the time since origin when the paint occurred, but the web vitals library may not report it until later. Same for FCP and TTFB.

packages/instrumentation-web-vitals/src/instrumentation.ts
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    this._listenersRegistered = true;
+
+    const trackingLevel = this.getConfig().trackingLevel ?? 'core';
+    this._diag.debug(
+      `Registering web vitals listeners (trackingLevel: ${trackingLevel})`,
+    );
+
+    onCLS((metric) => this._emitWebVital(metric));
```
martinkuba: Should we allow reporting all changes?
https://github.com/GoogleChrome/web-vitals#report-the-value-on-every-change

martinkuba (changes requested):

packages/instrumentation-web-vitals/src/instrumentation.ts
overbalance: Updated to include all

PR #162, opened 2026-02-17
feat: resource timing instrumentation

codecapitano:

Browser applications need visibility into resource loading performance to diagnose slow page loads, identify network bottlenecks, and optimize asset delivery. Currently, there's no standardized way to capture and emit resource timing data through OpenTelemetry in browser environments.

Adds a new `@opentelemetry/instrumentation-resource-timing` package that captures browser resource timing metrics for all loaded resources (scripts, stylesheets, images, XHR/fetch, fonts, etc.) and emits them as OpenTelemetry log events.

The implementation is designed to be non-intrusive:
- Uses `requestIdleCallback` to process resources during browser idle time, preventing interference with user interactions
- Includes a Safari-compatible shim that falls back to `setTimeout` when `requestIdleCallback` is unavailable
- Processes resources in batches (default: 50) to keep overhead predictable
- Implements queue limits and forced flushing to prevent memory issues during traffic spikes
- Flushes pending data when the page becomes hidden to prevent data loss during navigation

The idle callback shim is extracted into a separate module with dedicated tests, keeping browser compatibility concerns isolated from instrumentation logic.

## Type of change

## How Has This Been Tested?

Has unit tests and has been tested with a local running test web page.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "vitest": "4.1.2"
       },
       "peerDependencies": {
-        "typescript": "5.9.3"
+        "typescript": "6.0.2"
       }
     },
     "examples/all-instrumentations": {
@@ -6078,9 +6078,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.9.3",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.9.3.tgz",
-      "integrity": "sha512-jl1vZzPDinLr9eUt3J/t7V6FgNEw9QjvBPdysz9KfQDD41fQrC2Y4vKQdiaUpFT4bXlb1RHhLpp8wtm6M5TgSw==",
+      "version": "6.0.2",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-6.0.2.tgz",
+      "integrity": "sha512-bGdAIrZ0wiGDo5l8c++HWtbaNCWTS4UTv7RaTH/ThVIgjkveJt83m74bBHMJkuCbslY8ixgLBVZJIOiQlQTjfQ==",
       "license": "Apache-2.0",
       "peer": true,
       "bin": {
--- a/package.json
+++ b/package.json
@@ -55,7 +55,12 @@
     "vitest": "4.1.2"
   },
   "peerDependencies": {
-    "typescript": "5.9.3"
+    "typescript": "6.0.2"
+  },
+  "overrides": {
+    "typescript-eslint": {
+      "typescript": "^6"
+    }
   },
   "workspaces": [
     "examples/*",


packages/instrumentation-resource-timing/src/idle-callback-shim.ts
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const IDLE_DEADLINE_MS = 50;
+
+function supportsIdleCallback(): boolean {
+  // eslint-disable-next-line baseline-js/use-baseline
+  return typeof window.requestIdleCallback === 'function';
```
overbalance: nit: move this into the shim since it's only used once and add a comment why we find it useful enough to break the baseline rule. our of scope for this PR but moving this into common utils pkg after publish would be great

packages/instrumentation-resource-timing/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "build": "tsdown",
+    "watch": "tsdown --watch",
+    "check-types": "tsc",
+    "test": "vitest run --browser.enabled=false",
+    "test:watch": "vitest --browser.enabled=false",
+    "test:coverage": "vitest --coverage --browser.enabled=false"
+  },
+  "dependencies": {
+    "@opentelemetry/api-logs": "^0.211.0",
+    "@opentelemetry/instrumentation": "^0.211.0"
```
overbalance: just a little behind main
```suggestion
    "@opentelemetry/api-logs": "^0.212.0",
    "@opentelemetry/instrumentation": "^0.212.0"
```

packages/instrumentation-resource-timing/src/instrumentation.ts
```diff
@@ -0,0 +1,233 @@
...
+      document.removeEventListener(
+        'visibilitychange',
+        this._visibilityChangeHandler,
+      );
+      this._visibilityChangeHandler = undefined;
+    }
+  }
+
+  private _setupObserver(): void {
+    if (!this._isEnabled || !('PerformanceObserver' in window)) {
```
overbalance: we have put required checks like this in enable() since there is no other work that can be done if you're in a browser that has no PerformanceObserver. saves all the work that happens before this step.

thompson-tomo:
Agree with @overbalance we need a discussion about the purpose of the browser namespace in semconv and also we should be considering broader reusage of the telemetry.

Some quick thoughts:
- `browser.*` used to describe the platform/tool providing access to the Web based app
- `app.*` used to describe the website user is visiting including the blocks ie resources which make up the app
- `http.*` describes the transfer of resources from server to user

martinkuba:
I have made a few updates to this PR:
- moved the new instrumentation to the instrumentation package
- added it to the example (all-instrumentations)
- added configuration to filter based on initiatorType (for cases where there are too many resources)
- added baseline eslint exception for requestIdleCallback (instead of https://github.com/open-telemetry/opentelemetry-browser/pull/217)
- fixed issue with private fields (declare, so that they don't get initialized after base class touches them)

packages/instrumentation-resource-timing/src/instrumentation.ts
```diff
@@ -0,0 +1,233 @@
...
+        [ATTR_RESOURCE_DECODED_BODY_SIZE]: entry.decodedBodySize,
+        [ATTR_RESOURCE_REDIRECT_START]: entry.redirectStart,
+        [ATTR_RESOURCE_REDIRECT_END]: entry.redirectEnd,
+        [ATTR_RESOURCE_WORKER_START]: entry.workerStart,
+        [ATTR_RESOURCE_NEXT_HOP_PROTOCOL]: entry.nextHopProtocol,
+        [ATTR_RESOURCE_RENDER_BLOCKING_STATUS]: (
+          entry as PerformanceResourceTiming & {
+            renderBlockingStatus?: string;
+          }
+        ).renderBlockingStatus,
```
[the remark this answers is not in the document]
martinkuba: I think we should not include the attribute when there is no value.

Issue #131, opened 2026-01-23
Establish release and versioning process


martinkuba:
### Topic

We need to decide on a process for versioning packages and publishing them to npm.

### Decisions Needed

1. Versioning Strategy

    - Each package has its own version
    - All packages share the same version number

2. Version Bump Automation

    - Release Please, Changesets?
    - Manual

3. Release Workflow

    - Automatic, e.g. release PR is created/updated on every push to main, merge triggers publish
    - Manually trigger workflow to create release PR, separate workflow to publish

4. GitHub Releases

    - Create individual GitHub releases per package?
    - Or single release for all packages?

thompson-tomo:
My thoughts would be to go with the hybrid option but for naming I would go with:

- `@opentelemetry/instrumentation-browser` instrumentation which adds telemetry using browser api's
- `@opentelemetry/sdk-browser` implements functionality listed in the spec https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md and ideally they are also described in the spec
- `@opentelemetry/distro-browser` add's browser sdk/instrumentation and includes otlp exporter. It also configures defaults to help getting started.

I would version the 3 packages seperately and obviously each being seperate releases in github.

In relation to changelog, my suggestion would be to use a tool perhaps https://changie.dev/

joaquin-diaz:
I like the hybrid approach but I have two comments that I'd like to get your thoughts on. If we group all instrumentation that means they have shared dependencies, so:
1. Every time a dependency is updated, it needs to work in all the instrumentations. Instrumentations shouldn't depend on a lot of things, most likely only API and the base instrumentation package but something to have in mind
2. An instrumentation depending on an external library, let's say for example the web vitals library for the web vitals instrumentation would add this library to the entire instrumentations package. If user doesn't want to use the web vitals instrumentation, does this whole package get tree-shaken?

PR #175, opened 2026-03-05
build: consolidate instrumentations into a single package

martinkuba:

This is a follow-up to https://github.com/open-telemetry/opentelemetry-browser/issues/131

All existing instrumentation packages are going to be published in a single package in order to simplify the release process and compatibility matrix.

Added `browser-instrumentation` package, which contains the three existing instrumentations. They are all exported under the experimental subpath to signal that they are not stable.

The previous packages have been deleted. This is safe since they have not been published.


packages/browser-instrumentation/package.json
```diff
@@ -16,12 +15,13 @@
   "repository": {
     "type": "git",
     "url": "git+https://github.com/open-telemetry/opentelemetry-browser.git",
-    "directory": "packages/instrumentation-navigation-timing"
+    "directory": "packages/browser-instrumentation"
```
overbalance: I prefer `packages/instrumentation` as the folder name since we are already in the browser repo.

packages/browser-instrumentation/tsdown.config.ts
```diff
@@ -3,4 +3,10 @@ import baseConfig from '../../tsdown.config.ts';
 
 export default defineConfig({
   ...baseConfig,
+  entry: [
+    'src/navigation-timing/index.ts',
+    'src/user-action/index.ts',
+    'src/web-vitals/index.ts',
+  ],
```
overbalance: A wildcard would be less maintenance

PR #217, opened 2026-03-25
fix: allow requestIdleCallback in resource-timing dist baseline check

codecapitano:

The `instrumentation-resource-timing` package intentionally uses `requestIdleCallback` (not Baseline widely available — no Safari support), with browser compatibility handled via an `idle-callback-shim.ts` setTimeout fallback. The dist-level ESLint baseline check (`check:eslint:dist`) runs on compiled output where `eslint-disable` comments do not survive bundling, causing false positive failures in CI.

Adds a scoped override in `eslint.dist.config.js` that ignores the `requestidlecallback` feature ID for the `instrumentation-resource-timing` dist output only. All other packages remain fully covered by the baseline check.

## Type of change

## How Has This Been Tested?

`npm run check:eslint:dist` passes locally.

## Checklist:

### DIFF
--- a/eslint.dist.config.js
+++ b/eslint.dist.config.js
@@ -19,4 +19,20 @@ export default [
       ],
     },
   },
+  // resource-timing intentionally uses requestIdleCallback (not widely available)
+  // with a Safari fallback shim — suppress the dist-level check for this package.
+  {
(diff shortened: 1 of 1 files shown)


martinkuba:
This is now included in https://github.com/open-telemetry/opentelemetry-browser/pull/162

PR #218, opened 2026-03-26
docs: updated example in readme and added with-tracing example

martinkuba:

The README does not have an example relevant to what is in this repository.

Updated with a basic example for setting up instrumentations from this repository, and also added a new `with-tracing` example app that combines event-based and span-based instrumentations.

### DIFF
--- a/examples/with-tracing/.gitignore
+++ b/examples/with-tracing/.gitignore
@@ -1,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pnpm-debug.log*
+lerna-debug.log*
+
+node_modules
+dist
+dist-ssr
+*.local
+
+# Editor directories and files
+.vscode/*
+!.vscode/extensions.json
+.idea
+.DS_Store
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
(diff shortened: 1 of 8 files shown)


examples/with-tracing/.gitignore
overbalance, line 1: I would move the necessary entries to root and delete this file for consistency

examples/with-tracing/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    /* Linting */
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "erasableSyntaxOnly": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUncheckedSideEffectImports": true
+  },
+  "include": ["src"]
+}
```
overbalance, line 26:
```suggestion
{
  "extends": "../../tsconfig.json",
  "include": ["src"]
}
```

overbalance (approved):
2 nits for simplification but 👌

david-luna (approved):

examples/with-tracing/tsconfig.json
martinkuba: For the compilerOptions, the example needs `"types": ["vite/client"]`. Also, the existing example is already like this. If you want to simplify, can we handle it in a separate PR and apply to both examples?

PR #224, opened 2026-03-31
feat: add @opentelemetry/browser package

david-luna:

Add a new package named `@opentelemetry/browser` which hold the setup of the different signals separately and a bulder to compose them together into an SDK interface with start and stop methods.

This approach has taken into consideration that configuring all signals into a single SDK class presents challenges when it comes to tree shaking and bundlers cannot anticipate which modules are going to be used. So the configuration has been split by signal and needs to be imported explicitly by the consumer making easier for bundlers to analyse the code and remove unused modules.

You could start an SDK in isolation:
```ts
import {startLogsSdk} from '@opentelemetry/browser/logs';

const logsSdk = startLogsSdk({
  // configuration for logs
});

// then you can shutdown
await logsSdk.shutdown();
```

O combine more than one SDK

```ts
import {startLogsSdk} from '@opentelemetry/browser/logs';
import {startTracesSdk} from '@opentelemetry/browser/traces';

// Combine the functions into one
const startMySdk = combineSdks({
  logs: startLogsSdk,
  traces: startTracesSdk
});

// Use it to start
const mySdk = startMySdk({
  // It accepts global config options
  otlpEndpoint: 'http://my.collector.com'
  // And signal specific ones
  logs: {
    otlpLogsHeaders: { foo: 'bar' }
  }
});

// then you can shutdown
await mySdk.shutdown();
```

Some bundlers are smart enough to tree shake even if the SDKs are exported from a barrel file but explicit export paths has been added to be in sync with the approach done for instrumentations.

Refs #131 

## Type of change

## How Has This Been Tested?

- [ ] Unit test for the package
- [ ] Bundler tests to verify tree shaking is working

## Checklist:


david-luna:
@martinkuba 

I've changed to functions and removed the metrics. Still I have implemented a function to let the user combine logs & traces SDKs into one function. We can discuss all together the design in the next SIG

packages/browser/src/types.ts
```diff
@@ -0,0 +1,58 @@
...
+  disabled?: boolean;
+  logLevel?: DiagLogLevel;
+  // Resource & Entities related
+  serviceName?: string;
+  resource?: Resource;
+  // Export
+  otlpEndpoint?: string;
+  otlpHeaders?: Record<string, string>;
+  // Limits
+  attrLenghtLimit?: number;
```
overbalance:
```suggestion
  attrLengthLimit?: number;
```

packages/browser/src/logs.ts
```diff
@@ -0,0 +1,47 @@
...
+import { logs } from '@opentelemetry/api-logs';
+import { OTLPLogExporter } from '@opentelemetry/exporter-logs-otlp-http';
+import {
+  BatchLogRecordProcessor,
+  LoggerProvider,
+} from '@opentelemetry/sdk-logs';
+
+import type { LogsConfig, WebSdk } from './types.ts';
+
+const DEFAULT_LOGS_OTLP_ENDOINT = 'http://localhost:4318/v1/logs';
```
overbalance: I see a few of these:
```suggestion
const DEFAULT_LOGS_OTLP_ENDPOINT = 'http://localhost:4318/v1/logs';
```

packages/browser/src/traces.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http';
+import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
+import { WebTracerProvider } from '@opentelemetry/sdk-trace-web';
+
+import type { TracesConfig, WebSdk } from './types.ts';
+
+const DEFAULT_TRACES_OTLP_ENDOINT = 'http://localhost:4318/v1/metrics';
```
overbalance:
```suggestion
const DEFAULT_TRACES_OTLP_ENDOINT = 'http://localhost:4318/v1/traces';
```

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,72 @@
...
+        tracesConfig.otlpTracesEndpoint = otlpUrl.href;
+      }
+      tracesConfig.otlpTracesHeaders ??= globalConfig.otlpHeaders;
+      tracesConfig.resource ??= globalConfig.resource;
+      sdks.push(factories.traces(tracesConfig));
+    }
+
+    return {
+      shutdown() {
+        return Promise.all(sdks.map((s) => s.shutdown())).then(() => undefined);
```
overbalance:
```suggestion
        return Promise.allSettled(sdks.map((s) => s.shutdown())).then(() => undefined);
```
so the first failure doesn't short circuit flushing other instances

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,72 @@
...
+    globalConfig.resource ??= defaultResource();
+
+    const sdks: WebSdk[] = [];
+    const otlpUrl = new URL(globalConfig.otlpEndpoint);
+
+    // Start logs
+    if (factories.logs) {
+      const logsConfig = (config?.logs || {}) as LogsConfig;
+      if (!logsConfig.otlpLogsEndpoint) {
+        otlpUrl.pathname = 'v1/logs';
```
overbalance:
```suggestion
        otlpUrl.pathname = '/v1/logs';
```

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,72 @@
...
+      logsConfig.otlpLogsHeaders ??= globalConfig.otlpHeaders;
+      logsConfig.resource ??= globalConfig.resource;
+      sdks.push(factories.logs(logsConfig));
+    }
+
+    // Start traces
+    if (factories.traces) {
+      const tracesConfig = (config?.traces || {}) as TracesConfig;
+      if (!tracesConfig.otlpTracesEndpoint) {
+        otlpUrl.pathname = 'v1/traces';
```
overbalance:
```suggestion
        otlpUrl.pathname = '/v1/traces';
```

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type ConfigsFor<T> = Partial<{
+  [K in keyof T]: T[K] extends WebSdkFactory<infer C> ? C : never;
+}>;
+
+const DEFAULT_OTLP_ENDOINT = 'http://localhost:4318';
+
+export function combineSdks<T extends SdkFactories>(
+  factories: T,
+): WebSdkFactory<ConfigsFor<T>> {
```
overbalance:
```suggestion
): WebSdkFactory<GlobalConfig &
  ConfigsFor<T>> {
```

packages/browser/src/types.ts
```diff
@@ -0,0 +1,75 @@
...
+  // 1. `sampleRate` becomes a TraceIdRatioBasedSampler for traces
+  //    and maybe somethign else for other signals??? (sampling logs?)
+  // sampleRate?: number;
+}
+
+export interface LogsConfig {
+  // Resource & Entities related
+  resource?: Resource;
+  // Processor
+  blrpScheduleDelay?: number;
```
joaquin-diaz: Would it make sense to have an object grouping these options? Something like
```
...
batchLogRecordProcessorConfig: {
  scheduleDelay?: number;
  ..etc,
}
```

Same with traces config

david-luna: I think it does. Even we can do the same with the export config
```
exportConfig: {
  endpoint?: string;
  headers?: Record<string, string>;
}
```

but maybe there is no need to use such a long name. If used directly with the logs SDK the name of the method already gives the context

```js
startLogsSdk({
  // other config...
  processorConfig: {
    scheduleDelay: 5000,
  }
});
```

and if used with the compose API is nested in the the `logs` key

```js
// Combine the functions into one
const startMySdk = combineSdks({
  logs: startLogsSdk,
  traces: startTracesSdk
});

startMySdk({
  // other config...
  logs: {
    processorConfig: {
      scheduleDelay: 5000,
    }
  }
});
```

joaquin-diaz: Yeah, looks good

martinkuba:
I like the approach for making it tree-shakeable. In addition to that, I would like to propose a simple function like below for the most common use case (logs + traces + our recommended defaults)

```js
startBrowserSdk({
  serviceName: 'my-app',
  serviceVersion: '1.0.0',
  exportConfig: { url: 'https://collector' } },
  instrumentations: [new FetchInstrumentation(), new UserActionInstrumentation()],
});
```
This could just call combineSdks behind the scenes.

david-luna:
I'm going to close this PR and create a new one targeting a different branch https://github.com/open-telemetry/opentelemetry-browser/tree/feature/sdk-package

The motivation behind it is to allow implement this new package in smaller PRs and hopefully for more than one contributor. The package aims to contain all the necessary components to wireup and start an SDK so I guess there are going to be several discussions about different topics. So a divide and conquer strategy seems better to move on.

@martinkuba thanks for the feedback and I'm going to tackle https://github.com/open-telemetry/opentelemetry-browser/pull/224#issuecomment-4491556702 in the next PR :)

PR #225, opened 2026-03-31
feat(sandbox): introduce sandbox local working environment

mquentin:
# Note about this PR review

Changes of the `package-lock.json` make a 900lines of change to 3.8k lines of change. I know it can be difficult to review so I am up to any suggestion to ease the change review.

**This said, the PR can be reviewed by commit given that the package-lock.json changes are only part of the last commit**

The `opentelemetry-browser` repo has examples but does not have a self-hosted and simple sandbox that showcases the main browser OTel instrumentation.
This PR introduces this sandbox that you can start by running `npm run dev` and have a local web app instrumenting the OTel browser SDK.
This sandbox is also built with a view to being hosted on the GitHub Pages site of the repository.
# Note for the next step
This PR introduces the Sandbox but does not document it in the main `README.md` of the repo. A follow-up PR will add more info about it in the main readme.

## Changes

Introduce a `/sandbox` web app that initializes and configures the `opentelemetry-browser` with all the main browser instrumentations:

Screenshot 2026-04-01 at 16 23 59

### Sandbox changes

Introduce all the sandbox minimal code that enables running a simplistic app that initializes the OTel browser SDK, configures it, and adds the main supported browser instrumentations.
This includes `sandbox/src/otel.js` which contains the actual OTel logic and registers all instrumentations.

### Core changes

- [update] package.json: added sandbox to workspaces, npm run dev now starts the sandbox (old behavior preserved as npm run dev:example)
- [add] sandbox/vite.config.js: sets base for GitHub Pages in CI
- [add] .github/workflows/deploy-sandbox.yml: deploys sandbox/dist/ to GitHub Pages on push to main


package.json
```diff
@@ -59,7 +60,8 @@
   },
   "workspaces": [
     "examples/*",
-    "packages/*"
+    "packages/*",
+    "sandbox"
```
overbalance:
```suggestion
    "packages/*"
```

The sandbox should have its own lockfile so the devDeps don't pollute the core source code

package.json
```diff
@@ -19,7 +19,8 @@
     "prepare": "lefthook install -f",
     "generate": "turbo generate",
     "clean": "npx -y rimraf -g \"**/.turbo\" \"**/dist\" \"**/node_modules\"",
-    "dev": "turbo run watch dev --filter=examples-all-instrumentations...",
+    "dev": "turbo run watch dev --filter=sandbox...",
+    "dev:example": "turbo run watch dev --filter=examples-all-instrumentations...",
```
overbalance:
```suggestion
    "dev": "npm i -p sandbox && npm run dev -p sandbox",
```
After removing sandbox from workspaces, this dev command becomes a convenience method

sandbox/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@opentelemetry/exporter-trace-otlp-http": "^0.214.0",
+    "@opentelemetry/instrumentation": "^0.214.0",
+    "@opentelemetry/instrumentation-fetch": "^0.214.0",
+    "@opentelemetry/instrumentation-xml-http-request": "^0.214.0",
+    "@opentelemetry/resources": "^2.0.0",
+    "@opentelemetry/sdk-logs": "^0.214.0",
+    "@opentelemetry/sdk-trace-base": "^2.0.0",
+    "@opentelemetry/sdk-trace-web": "^2.0.0",
+    "@opentelemetry/semantic-conventions": "^1.40.0",
+    "@picocss/pico": "^2.0.0",
```
overbalance: This is pulling in a ton of old dependencies. Can you try including `https://cdnjs.cloudflare.com/ajax/libs/picocss/2.1.1/pico.min.css` in the html, instead of importing the node module?

sandbox/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
```
overbalance: We should make sure this is > 19.2.0
https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components
```suggestion
    "react": "^19.2.1",
    "react-dom": "^19.2.1"
```

sandbox/.gitignore
overbalance, line 1: You can safely delete this file

sandbox/src/main.jsx
```diff
@@ -0,0 +1,6 @@
+import '@picocss/pico/css/pico.min.css'
```
overbalance: See my comment about pico

package.json
overbalance:
```suggestion
    "example": "turbo run watch dev --filter=examples-all-instrumentations...",
```
I'd keep this simple since it's not actually part of the `dev` command namespace like `lint` and `lint:fix`

.github/workflows/deploy-sandbox.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches: [main]
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: true
```
overbalance: I believe you need this off so failed checks never produce incomplete uploads

sandbox/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@opentelemetry/exporter-logs-otlp-http": "^0.214.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.214.0",
+    "@opentelemetry/instrumentation": "^0.214.0",
+    "@opentelemetry/instrumentation-fetch": "^0.214.0",
+    "@opentelemetry/instrumentation-xml-http-request": "^0.214.0",
+    "@opentelemetry/resources": "^2.0.0",
+    "@opentelemetry/sdk-logs": "^0.214.0",
+    "@opentelemetry/sdk-trace-base": "^2.0.0",
+    "@opentelemetry/sdk-trace-web": "^2.0.0",
+    "@opentelemetry/semantic-conventions": "^1.40.0",
```
overbalance, line 25: Make sure these are identical to the root before merge so there are not mixed versions. You can check the lockfile. The dupes are obvious because you'll see an OTel package inside another OTel package with slightly different versions

sandbox/README.md
```diff
@@ -0,0 +1,22 @@
+# Sandbox
+
+Interactive playground for testing OpenTelemetry browser instrumentations. Configure the SDK, trigger actions (fetch, XHR, errors, navigation…), and inspect exported spans and logs in real time.
+
+## Run locally
+
+From the **repo root**:
```
overbalance: This is no longer true if you accept my suggestions. You can run install and dev from root, or inside the sandbox folder

sandbox/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
+    "jsx": "react-jsx",
+    "skipLibCheck": true,
+    "moduleResolution": "bundler",
+    "allowJs": true,
+    "checkJs": false,
+    "noEmit": true,
+    "strict": true,
+    "erasableSyntaxOnly": true
+  },
```
overbalance, line 5:
```suggestion
  "extends": "../tsconfig.json",
  "compilerOptions": {
    "jsx": "react-jsx",
  },
```

sandbox/vite.config.js
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vite'
+import react from '@vitejs/plugin-react'
+
+export default defineConfig({
+  plugins: [react()],
```
overbalance: I think vite automatically handles jsx now. Try it without this plugin.

sandbox/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@picocss/pico": "^2.0.0",
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-react": "^4.5.0"
```
overbalance: See my comment about vite

overbalance (changes requested):

joaquin-diaz:
Any chance of using typescript instead of plain javascript?

sandbox/.gitignore
mquentin: Indeed covered by the parent .gitignore thanks

sandbox/vite.config.js
mquentin: Indeed it works perfectly without this plugin. Great suggestion

sandbox/package.json
mquentin: Indeed that's what was bringing so much package-lock.json file diffs. I did not realize the examples were using other packages versions

mquentin: Thanks for helping me on this

mquentin:
> Any chance of using typescript instead of plain javascript?

Hey, great suggestion, I went for this change too.

sandbox/index.html
```diff
@@ -0,0 +1,16 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>OTel Browser Instrumentation Sandbox</title>
+    <link
+      rel="stylesheet"
+      href="https://cdnjs.cloudflare.com/ajax/libs/picocss/2.1.1/pico.min.css"
+    >
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="./src/main.jsx"></script>
```
overbalance:
```suggestion
    <script type="module" src="./src/main.tsx"></script>
```

sandbox/src/utils/config.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function parseJson(
+  raw: string | null,
+  fallback: Record<string, string>,
+): Record<string, string> {
+  if (!raw) {
+    return fallback;
+  }
+  try {
+    const parsed: unknown = JSON.parse(decodeURIComponent(raw));
```
overbalance:
```suggestion
    const parsed = JSON.parse(decodeURIComponent(raw));
```

sandbox/src/app/actions.ts
```diff
@@ -0,0 +1,159 @@
...
+    const url = 'https://jsonplaceholder.typicode.com/users/1';
+    const req = new XMLHttpRequest();
+    req.open('GET', url);
+    req.send();
+  };
+
+  const jsError = () => {
+    const errorSpan = tracer.startSpan('js-error-event');
+    try {
+      void (null as unknown as { undefinedProperty: string }).undefinedProperty;
```
overbalance:
```suggestion
    // @ts-expect-error generating an error on purpose
    void {}.undefinedProperty;
```

sandbox/src/app/App.tsx
```diff
@@ -0,0 +1,368 @@
...
+    cfg.initial.serviceVersion,
+    cfg.initial.tracesUrl,
+  ]);
+
+  // ── Auto-scroll log ───────────────────────────────────────────────────────
+  useEffect(() => {
+    if (logBodyRef.current) {
+      logBodyRef.current.scrollTop = logBodyRef.current.scrollHeight;
+    }
+  }, []);
```
overbalance: I think this only runs once without a dependency
```suggestion
  }, [logs]);
```

sandbox/src/utils/config.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function readConfigFromForm(): ParsedConfig {
+  return {
+    serviceName: field('ub-sn') || DEFAULTS.serviceName,
+    serviceVersion: field('ub-sv') || DEFAULTS.serviceVersion,
+    tracesUrl: field('ub-traces-url') || DEFAULTS.tracesUrl,
+    logsUrl: field('ub-logs-url') || DEFAULTS.logsUrl,
+    customAttributes: {}, // populated separately via readCustomAttributes()
+  };
+}
```
overbalance: Are these used?

sandbox/src/utils/config.ts
```diff
@@ -0,0 +1,77 @@
...
+    const parsed: unknown = JSON.parse(decodeURIComponent(raw));
+    if (
+      typeof parsed === 'object' &&
+      parsed !== null &&
+      !Array.isArray(parsed)
+    ) {
+      return parsed as Record<string, string>;
+    }
+  } catch {
+    /* ignore */
```
overbalance: Let users diagnose malformed url attrs
```suggestion
  } catch (e) {
    console.error(e);
```

sandbox/src/app/App.tsx
```diff
@@ -0,0 +1,368 @@
...
+      serviceVersion: cfg.initial.serviceVersion,
+      tracesUrl: cfg.initial.tracesUrl,
+      logsUrl: cfg.initial.logsUrl,
+    };
+    const attrs = { ...cfg.initial.customAttributes };
+
+    try {
+      const handle = initOtel(config, attrs, {
+        onSpan: (type, msg) => addLog(type, msg),
+        onLog: (type, msg) => addLog(type, msg),
```
overbalance:
```suggestion
        onSpan: addLog,
        onLog: addLog,
```

sandbox/src/app/App.tsx
```diff
@@ -0,0 +1,368 @@
...
+  ]);
+
+  // ── Auto-scroll log ───────────────────────────────────────────────────────
+  useEffect(() => {
+    if (logBodyRef.current) {
+      logBodyRef.current.scrollTop = logBodyRef.current.scrollHeight;
+    }
+  }, []);
+
+  // ── Syntax-highlighted code snippet (avoids dangerouslySetInnerHTML) ──────
```
overbalance: Why does it need to be avoided? Is innerHTML not exactly the same?

sandbox/src/app/actions.ts
```diff
@@ -0,0 +1,159 @@
...
+    for (const [i, name] of stepNames.entries()) {
+      const step = tracer.startSpan(`workflow.step-${i + 1}`);
+      step.setAttribute('step.index', i + 1);
+      step.setAttribute('step.name', name);
+      await sleep(50 + (i + 1) * 30);
+      step.end();
+    }
+
+    root.end();
+  };
```
overbalance, line 115: As written, I think the spans are siblings, not nested.

```suggestion
  const nestedSpans = async () => {
    const root = tracer.startSpan('workflow.execute');
    root.setAttribute('workflow.name', 'demo-pipeline');
    root.setAttribute('workflow.steps', 3);

    await sleep(40);

    const rootCtx = trace.setSpan(context.active(), root);
    const stepNames = ['validate', 'process', 'commit'];
    for (const [i, name] of stepNames.entries()) {
      await context.with(rootCtx, async () => {
        const step = tracer.startSpan(`workflow.step-${i + 1}`);
        step.setAttribute('step.index', i + 1);
        step.setAttribute('step.name', name);
        await sleep(50 + (i + 1) * 30);
        step.end();
      });
    }

    root.end();
  };
```

overbalance (changes requested):

sandbox/src/app/App.tsx
mquentin: So this `useCallback` does not need a `logs` dependency as `setLogs` mutate its previous state by itself by appending a new entry `{ id, type, msg, time }`.

One could have written it like:

```
  const addLog = useCallback((type: string, msg: string) => {
    const id = ++logIdRef.current;
    const time = new Date().toLocaleTimeString([], {
      hour: '2-digit',
      minute: '2-digit',
      second: '2-digit',
    });
    setLogs([...logs, { id, type, msg, time }]);
  }, [logs]);
```

but it means rememoizing the callback after each `logs` change.

Like you mention with the current version we can afford running the memoization only once thanks to the `setLogs((prev) => [...prev, `

sandbox/src/app/actions.ts
mquentin: Thanks for the change suggestion ! My bad, I mainly wrote it to have some span example data, it's even better if it makes sens and actually generate nested sets of span

sandbox/src/utils/config.ts
mquentin: Indeed, you're right I don't use this code anymore since I went for a React approach. Thanks

sandbox/src/app/App.tsx
mquentin: I think it's also a dead code that remains from my vanilla JS implementation where I had to escape manually the user input content before building the code snippet. React already deals with all this for us now. Thanks for spotting it.

mquentin: I will clean that

mquentin:
@overbalance Thanks for the review. On top of you suggestion, I also refactored the code structure a bit so we have clearer React components, hooks structure as well as a better typing. Let me know what you think about this

sandbox/src/utils/sdk-config-parser.ts
overbalance, line 1: Please rename to sdk-config-parser.ts

sandbox/package.json
overbalance: You'll have to do this one more time because the versions were updated by renovate. Thanks for all your work on this

sandbox/src/utils/sdk-config-parser.ts
mquentin: great suggestion, pushed the change

sandbox/package.json
mquentin: Good point, as the `examples/all-instrumentations/package.json` were also not up to date with the root with the latest renovate changes I took the liberty to update its version too.

sandbox/package.json
overbalance: And remove it from here please

overbalance (approved):

Issue #174, opened 2026-03-03
Instrument browser app screen url


mquentin:
### Problem statement

So I am looking for a specific OTel RUM instrumentation: **I need to know on which screen/page a click interaction happened**.

For now we have `app.screen.name` and `app.screen.id` which is great as it is agnostic of any browser and mobile context. However for a browser app, I do not see yet a proper way of instrumenting the `location.href` of my screen.

In the example of a click interaction, I cannot leverage the [url.full](https://github.com/open-telemetry/semantic-conventions/blob/c3485c76776a2a6cdfbfc43dae5ab0e1956bb10a/model/url/registry.yaml#L39) to hold my `location.href` as my click did not trigger any network call and therefore the `url.full: Absolute URL describing a network resource` does not make sense.

### Suggested solution

I am suggesting we introduce a new `app.screen.url.full` field. This field can also match a Mobile context given that a lot of Mobile apps contain webviews and therefore can also have a `screen.url`.

Like this one can have a `app.screen.name: Login` and also have a more specific field such as `app.screen.url.full:www.myfullurl.com/path?param1...`

This would also enable some url browser processor such as https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/web-common/src/SessionLogRecordProcessor.ts that would automatically populate the `app.screen.url.full` with the browser API `location.href` without requiring users to configure any `app.screen.name`.

### Alternative solution

An alternative would be to update the definition of [url.full](https://github.com/open-telemetry/semantic-conventions/blob/c3485c76776a2a6cdfbfc43dae5ab0e1956bb10a/model/url/registry.yaml#L39) to be more permissive and allow holding network or location urls.

mquentin:
# 16th of April - Browser SIG

To reduce confusion with other platform while complying the the Browser API: https://developer.mozilla.org/en-US/docs/Web/API/Document/URL, the SIG decided to go for `browser.document.url.full` and start a POC in the https://github.com/open-telemetry/opentelemetry-browser sandbox to instrument it

PR #237, opened 2026-04-22
feat:instrument browser document url full

mquentin:

Stacked on top
   of https://github.com/open-telemetry/opentelemetry-browser/pull/225 and https://github.com/open-telemetry/semantic-conventions/pull/3633 providing a solution to this issue: https://github.com/open-telemetry/opentelemetry-browser/issues/174#issuecomment-4091336306

Fixes # https://github.com/open-telemetry/opentelemetry-browser/issues/174#issuecomment-4091336306

Adds `BrowserDocumentUrlInstrumentation` that stamps every OTel span and every OTel log record event with `browser.document.url.full` derived from `location.href` at creation time.

Also add this new instrumentation to the sandbox testing environment

Screenshot 2026-04-22 at 11 55 21
Screenshot 2026-04-22 at 11 46 50

# Note - comment about alternative approaches:

We just add a log/span processor classes that add this attribute to all logs/spans. There is already a precedent for session.id [here](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/web-common/README.md#sessions).
We work towards making this a resource attribute.

The second option is ultimately what we want (for both this and session). However, the mechanism for this is not defined yet. My hope is to prototype this in the end-to-end demo that we have discussed. And I would like to prototype it in a way that it can support multiple different attributes, using entities, and show how updates would be handled by the SDK. I described this in the [discussion topic](https://github.com/open-telemetry/opentelemetry-browser/discussions/208).

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## How Has This Been Tested?

- [ ] Test A

## Checklist:


packages/instrumentation/src/browser-document-url/instrumentation.ts
joaquin-diaz, line 1: Is this an approach we do for other instrumentations? I'm wondering if instead of trying to patch the existing logger provider we should instead just have a log processors package that users can use when they init the SDK. I don't like that we're depending on an internal API (_activeSpanProcessor) to add the processor

mquentin: So this is really a draft of a draft. I am actually up to any suggestion because I started this more as a demo than a PR proposal.

martinkuba:
This was implemented in #421. Closing.

Issue #238, opened 2026-04-22
Move instrumentation-browser-navigation from contrib to this repo


martinkuba:
Sub-issue of #209.

Move `@opentelemetry/instrumentation-browser-navigation` from [opentelemetry-js-contrib](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-browser-navigation) into the `@opentelemetry/browser-instrumentation` package as a subpath export.

## Checklist

- [x] Move code to this repo
	- [x] Add instrumentation code to `packages/instrumentation/src/navigation/`
	- [x] Add subpath export (`./experimental/navigation`) to `package.json`
	- [x] Adapt code to repo conventions (logs-based event API, vitest, tsdown build, shared test utils)
	- [x] Add/port tests
- [ ] Deprecate the js-contrib package 
	- [ ] Update the `browser-instrumentation` README with docs for the new instrumentation
	- [ ] Update the contrib package README with deprecation notice and migration instructions, publish a final version
	- [ ] Run `npm deprecate` on `@opentelemetry/instrumentation-browser-navigation` pointing to the new location
	- [ ] Delete `instrumentation-browser-navigation` package source from opentelemetry-js-contrib

PR #242, opened 2026-04-23
feat(instrumentation): port browser-navigation into consolidated package

martinkuba:
## Summary

Ports `instrumentation-browser-navigation` from opentelemetry-js-contrib into `@opentelemetry/browser-instrumentation` under a new `./experimental/navigation` subpath. Addresses the first checklist item of #238 ("Move code to this repo"); deprecation/deletion of the contrib package is tracked as separate items in that issue.

Adapted to repo conventions: `InstrumentationBase` + `this.logger.emit()`, `declare` field pattern, `safeExecuteInTheMiddle` around the user hook, `verbatimModuleSyntax`-friendly imports, vitest tests via `setupTestLogExporter`.

## Test plan

- [x] `npm run check-types` — clean
- [x] `npm run build` — clean, including `attw` and `publint`
- [x] `npm test` — 119/119 passing (unit + browser projects)
- [x] Repo-root `turbo build` — successful, examples build picks up the new subpath export

Refs #238

### DIFF
--- a/packages/instrumentation/src/navigation/utils.ts
+++ b/packages/instrumentation/src/navigation/utils.ts
@@ -1,0 +1,105 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const SENSITIVE_PARAMS = [
+  'password',
+  'passwd',
+  'secret',
+  'api_key',
+  'apikey',
+  'auth',
+  'authorization',
+  'token',
+  'access_token',
+  'refresh_token',
+  'jwt',
+  'session',
+  'sessionid',
+  'key',
+  'private_key',
+  'client_secret',
+  'client_id',
+  'signature',
+  'hash',
+];
+
+/**
+ * Default URL sanitization function that redacts credentials and sensitive query parameters.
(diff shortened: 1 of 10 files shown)


joaquin-diaz (approved):

packages/instrumentation/src/navigation/utils.ts
```diff
@@ -0,0 +1,105 @@
+/*
```
david-luna, line 1: question: have you thought about placing these functions in https://github.com/open-telemetry/opentelemetry-browser/tree/main/packages/instrumentation/src/utils ?

david-luna (approved):

packages/instrumentation/src/navigation/utils.ts
martinkuba: It makes sense for at least `defaultSanitizeUrl` - it's a generic URL sanitizer that any URL-capturing instrumentation (e.g. resource-timing, future fetch/XHR) could reuse. `isHashChange` is more navigation-specific.

Since this PR is scoped as a straight port from contrib, let's handle the consolidation in a follow-up that can survey shared helpers across all instrumentations. Tracking it in #251.

PR #248, opened 2026-04-27
fix(resource-timing): prevent feedback loop by adding ignoreUrls

mquentin:

The resource timing instrumentation does not have ignored URL API like the other core instrumentation (cf. XMLHttpRequestInstrumentation, FetchInstrumentation)

As a result, when the SDK post the signals to intakes, it is caught as a resource which has a timing which create a new signal to post to intake...

This PR adds an ignoreURL API to the `ResourceTimingInstrumentation` implemented like the XMLHttpRequestInstrumentation one.

I also configure this existing API for XMLHttpRequestInstrumentation, FetchInstrumentation and this new API for ResourceTimingInstrumentation in the Sandbox env to prevent infinite loop to happens.

### Before

https://github.com/user-attachments/assets/b279948c-0aa8-4fe4-9553-e4744f119f96

### After

https://github.com/user-attachments/assets/eda7a8be-4a56-43d1-ad0f-2cd4ad7100f9

## Type of change

## How Has This Been Tested?

- [ ] Test A

## Checklist:


packages/instrumentation/src/resource-timing/instrumentation.ts
```diff
@@ -133,44 +134,50 @@ export class ResourceTimingInstrumentation extends InstrumentationBase<ResourceT
...
+
+    this._observer = observer;
+
+    try {
       observer.observe({ type: 'resource', buffered: true });
     } catch {
       this._diag.warn(
         'PerformanceObserver not supported, resource timings will not be collected',
       );
+      this._observer = undefined;
```
overbalance: Since this is narrowed, consider making it more specific:
```suggestion
    } catch (e) {
      this._diag.warn(
        'Failed to start resource PerformanceObserver', e
      );
      this._observer = undefined;
```

packages/instrumentation/src/resource-timing/types.ts
```diff
@@ -41,4 +41,12 @@ export interface ResourceTimingInstrumentationConfig
    * are captured. When unset, all resource entries are captured.
    */
   initiatorTypes?: string[];
+
+  /**
+   * URLs to ignore. Entries whose URL matches any of the patterns will not be
+   * captured. Strings are matched exactly; RegExps are tested against the full URL.
```
overbalance:
```suggestion
   * captured. Strings are matched exactly; RegExps are matched against the full URL.
```

packages/instrumentation/src/resource-timing/types.ts
```diff
@@ -41,4 +41,12 @@ export interface ResourceTimingInstrumentationConfig
...
+   * Avoid RegExps with the `g` or `y` flag — their stateful `lastIndex` causes
```
overbalance: I think `y` here causes alternating matches, but not `g`

packages/instrumentation/src/resource-timing/types.ts
overbalance: This says "matched exactly" but I think we should be a bit louder. Because OTel uses a strict equality, this is case-sensitive and not normalized so a trailing slash or query string reintroduces the buggy loop. 

Maybe suggest an example like`[/\/v1\/traces$/, /\/v1\/logs$/]` at the bottom.

overbalance (changes requested):

packages/instrumentation/src/resource-timing/instrumentation.ts
martinkuba: We could also log the actual error (using the diagnostics logger).

packages/instrumentation/src/resource-timing/instrumentation.ts
```diff
@@ -133,44 +134,50 @@ export class ResourceTimingInstrumentation extends InstrumentationBase<ResourceT
...
-            !initiatorTypes.includes(entry.initiatorType)
-          ) {
-            continue;
-          }
-
-          if (this._pendingEntries.length >= maxQueueSize) {
-            this._flush();
-          }
-          this._pendingEntries.push(entry);
+        if (isUrlIgnored(entry.name, this._config.ignoreUrls)) {
```
martinkuba: If this throws because of a bad regex, then the instrumentation does not recover - `_scheduleProcessing` is never called, and also no log that this has happened. I would either wrap the `isUrlIgnored` in try/catch or at least log an error.

martinkuba (changes requested):

packages/instrumentation/src/resource-timing/instrumentation.ts
mquentin: Great suggestion, I will try catch this and leverage the diagnostics logger

martinkuba (approved):

overbalance (approved):

PR #252, opened 2026-04-28
feat: translate event_name from OTLP to event_name attribute in Loki

martinkuba:
This makes the event_name value available on logs in Loki.

- Update collector contrib to a version that reads the event_name field
- Add a transform that copies the event_name value into the event.name attribute


david-luna (approved):
LGTM

PR #281, opened 2026-05-15
feat(browser-instrumentation): add fetch instrumentation

### LINKED ISSUE #259: Move fetch instrumentation from JS core to this repo
Sub-issue of https://github.com/open-telemetry/opentelemetry-browser/issues/209

Move the `@opentelemetry/instrumentation-fetch` instrumentation from [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-fetch).

Related to  https://github.com/open-telemetry/opentelemetry-js/issues/6591 and https://github.com/open-telemetry/opentelemetry-js/pull/6630

Decisions needed
- should it be part of the consolidated instrumentation package or separate? (the current package is widely used)
- where should it import utils from (can be separate issue)
- should we keep the span events for resource timing or make it work with the new resource timing instrumentation (can be separate issue)

david-luna:

Closes #259

Inclusion of the `fetch `instrumentation with some differences from the original version in https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-fetch

- utils moved to this package instead of using `@opentelemetry/sdk-trace-web`
- removed dependency from `@opentelemetry/core`
- only emit span for the fetch call. it does not emit child spans for preflight requests neither span event for the resource timings

This PR uses a different approach than https://github.com/open-telemetry/opentelemetry-browser/pull/283 when it comes to resource timings. In here the fetch instrumentation stashes temporarily the context for the fetch request. this is done using a util for that purpose. Then the `resource-timing` instrumentation does a query to get the context related to the resource timing entry and, if found, attach the context to the log record.

So in order to get all data you need both instrumentations (fetch, resource-timings) enabled and make sure `ignoreUrls` configuration for both instrumentations is compatible. IMHO it simplifies the instrumentation implementation and reuses code that already works while avoiding duplication of data (logs for resource timings). The caveat is in the configuration but the alternative also need to be careful on config to avoid duplicated logs.

## Type of change

## How Has This Been Tested?

- [x] Unit tests for new utils
- [x] Browser tests for instrumentation

## Checklist:


packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,410 @@
...
+        let options: Request | RequestInit;
+        if (args[0] instanceof Request) {
+          options = args[1] != null ? new Request(args[0], args[1]) : args[0];
+        } else {
+          options = args[1] || {};
+        }
+
+        const createdSpan = instrumentation._createSpan(url, options);
+        if (instrConfig.measureRequestSize) {
+          getFetchBodyLength(...args)
```
joaquin-diaz: Can we use async/await here?

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ *   ReadableStream.pipeThrough() lets us process a stream and returns a new stream
+ *   So we can measure the body length as it passes through the pie, but need to attach
+ *   the new stream to the original request
+ *   so that the browser still has access to the body.
+ *
+ * @param body
+ * @returns promise that resolves to the content length of the body
+ */
+export function getFetchBodyLength(...args: Parameters<typeof fetch>) {
```
joaquin-diaz: Also let's try to use async here

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,410 @@
...
+            .then((bodyLength) => {
+              if (typeof bodyLength === 'number') {
```
joaquin-diaz: Can we type this correctly and return null | number instead of having to check this in here?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,410 @@
...
+              }
+            })
+            .catch((error) => {
+              instrumentation._diag.warn('getFetchBodyLength', error);
+            });
+        }
+
+        function endSpanOnError(span: Span, error: FetchError) {
+          instrumentation._applyAttributesAfterFetch(span, options, error);
+          instrumentation._endSpan(span, {
```
joaquin-diaz: I don't see where we use statusText and url in `_endSpan`, it most places we send it. Then in some places we do something like:
```
            instrumentation._endSpan(span, {
              status: response.status,
              statusText: response.statusText,
              url,
            });
```

where it would be the same as doing `instrumentation._endSpan(span, response)` since we don't use url. Am I missing something?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,410 @@
...
+
+            // Call request hook before injection so hooks cannot tamper with propagation headers.
+            // Also, this means the hook will see `options.headers` in the same type as passed in,
+            // rather than as a `Headers` instance set by `_addHeaders()`.
+            instrumentation._callRequestHook(createdSpan, options);
+            instrumentation._addHeaders(options, url);
+
+            return original
+              .apply(
+                // @ts-expect-error
```
joaquin-diaz: Any way that we can avoid these comments?

packages/instrumentation/src/utils/url.ts
```diff
@@ -54,3 +54,74 @@ export function defaultSanitizeUrl(url: string): string {
...
+}[keyof T];
+type UrlKeys = KeysOfType<URL, string>;
+export type URLLike = Pick<URL, UrlKeys>;
+
+const ANCHOR_ELEMENT = document.createElement('a');
+/**
+ * Parses url using URL constructor or fallback to anchor element.
+ * @param url
+ */
+export function parseUrl(url: string): URLLike {
```
joaquin-diaz: URL is Baseline Widely available I don't think we need the fallback anymore

david-luna:
@joaquin-diaz thanks for your review :)

I did not have a chance to do a detailed review of yours https://github.com/open-telemetry/opentelemetry-browser/pull/283 and still have no string opinion on which route we should take. I'll drop here my thoughts


I think the main difference in the approach in both PRs is where is the resource timings handled. We have 2 different approaches

- Timings handled by a single instrumentation. Other instrumentation may provide data/context for a given resource (what is done in this PR)
  - pros:
    - avoiding unnecessary observers in the client
    - is not possible to have duplicated logs for resource timing entries
    - singe responsibility principle. complexity of dealing with times is only in one place
  - cons:
    - Some config options become split between instrumentations. For example `ignoreUrls`. If resource timings instrumentation ignores a URL not ignored by fetch then the fetch Span won't have ant correlated logs
    - you need 2 instrumentations to have all the data (spans/logs)

- Timings handled by any instrumentation that needs the data
  - pros:
    - one instrumentation to get and export all data. 
    - no config split in 2 instrumentations
  - cons:
    - Config can be an issue too if you don't want to send duplicated logs. For example resource timings for a fetch request. Resource timings instrumentation should ignore the fetch entries already collected by the fetch instrumentation.
    - Possible duplication of logic/code. It can be mitigated moving some of the login into `utils`

Also there is the initiative to remove `HrTime` from browser instrumentations s I think we need to leverage this migration to already do that. ref: https://github.com/open-telemetry/opentelemetry-js/pull/6555

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: good question. I think `await`ing here is going to halt (virtually) the execution of this logic resulting in a degradation of fetch requests. So IMHO is better to not call `await` here.

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: Do you mean we should report always the fetch length? with the numeric value or null?

david-luna: looking at the [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/b8ff3eb9508d6699642b018a9dc19100244f5cc7/model/http/registry.yaml#L8) i see the attribute is defined as an `int`. So I think it makes sense the have the check and report only if the result is of type number.

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: This property was used in the original implementation [here](https://github.com/open-telemetry/opentelemetry-js/blob/6adb9b4f10f5f22d9802f71e6db20303d7ca4e3c/experimental/packages/opentelemetry-instrumentation-fetch/src/fetch.ts#L169). Now is not needed anymore. 

Good catch :)

david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/d8d8fd215e98ef3055d2ec0d18b2c0f527836763

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: in this specific case we can substitute it with `globalThis`. Done in
- https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/a6d03f2ad5c97ab65a88ce045b31026bfb9360ff
- https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/5080aabca1cbd4c353b7273f6e120da791536200

packages/instrumentation/src/utils/request.ts
david-luna: I agree that using `async` here can allow us to write this function in a more clear way.  However I think it shouldn't be awaited in the fetch instrumentation code.

packages/instrumentation/src/utils/url.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/a1f616150cb0be9e8862a541a2cbc0d78af75d7a

david-luna:
@joaquin-diaz 

I've updated this PR to use the `NetworkContextRegistry` from https://github.com/open-telemetry/opentelemetry-browser/pull/329. Now `fetch` and `resource-timing` use the registry to correlate the logs to the fetch spans.

Now is good for review but I was thinking to include the e2e tests you implemented in https://github.com/open-telemetry/opentelemetry-browser/pull/283. WDYT?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+          return original
+            .apply(
+              globalThis,
+              options instanceof Request ? [options] : [url, options],
+            )
+            .then(
+              onSuccess.bind(globalThis, createdSpan),
+              onError.bind(globalThis, createdSpan),
+            )
+            .finally(() => {
```
overbalance: This runs in `.finally`, but `onSuccess` only starts reading the body and returns right away. So `responseEnd` here is when the headers arrived, not when the body finished. The real `responseEnd` from the browser is later, so the check `entry.responseEnd <= r.endPerfNow` in `getContext` fails and correlation is lost for larger or streamed responses. Small responses pass because headers and body arrive almost together. What do you think about registering from `_endSpan` instead, when reading is really done?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+    // Add to the registry and keep a reference
+    registry.register(span, data);
+    this._registeredResources = this._registeredResources || [];
+    this._registeredResources.push(resource);
+
+    // Cancel any pending clear task and schedule
+    if (typeof this._unregisterTimer === 'number') {
+      clearTimeout(this._unregisterTimer);
+    }
+    this._unregisterTimer = setTimeout(() => {
```
overbalance: This timer is cleared and set again on every request. If requests keep coming less than 1s apart (polling, streaming), it never fires and `_registeredResources` grows forever. One timer per resource, or a periodic sweep, would avoid that. Also, if `unregister` throws, the loop stops before the reset on 306-307 and cleanup gets stuck. Moving the reset into a `finally` would prevent it.

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+            // from the original response. But it does work with `AbortController.abort()`
+            // because it aborts the underliying fetch cancelling the original and clone streams
+            // ref: https://github.com/open-telemetry/opentelemetry-js/pull/6521
+            const resClone = response.clone();
+            const body = resClone.body;
+            if (body) {
+              const reader = body.getReader();
+
+              const read = (): void => {
+                reader.read().then(
```
overbalance: If `endSpanOnSuccess` or `endSpanOnError` throws here, this promise rejects with no handler and the span never ends. `onError` below (231-243) already guards this. Same try/catch here would be safer:

```ts
({ done }) => {
  try {
    if (done) endSpanOnSuccess(span, response);
    else read();
  } catch (e) {
    instrumentation._diag.error('Failed to finalize span after body read', e);
    instrumentation._endSpan(span, { status: 0 });
  }
},
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+      return function patchConstructor(
+        this: typeof globalThis,
+        ...args: Parameters<typeof fetch>
+      ): Promise<Response> {
+        if (!instrumentation._isEnabled) {
+          return original.apply(this, args);
+        }
+        const instrConfig = instrumentation.getConfig();
+
+        const url = parseUrl(
```
overbalance: All of this (`parseUrl`, `_createSpan`, `_callRequestHook`, `_addHeaders`) runs before `original.apply`, with no try/catch. `parseUrl` can throw on a bad URL, or when there is no `document`/`location` (workers). If it throws, the user's fetch throws too. Could we wrap it and fall back to `original.apply(this, args)` on error?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+  ): Span {
+    const attributes = {} as Attributes;
+    const parsedUrl = parseUrl(url);
+    const origMethod = options.method;
+    const normMethod = normalizeHttpRequestMethod(options.method || 'GET');
+    const serverPort = serverPortFromUrl(parsedUrl);
+    const name = normMethod;
+
+    attributes[ATTR_HTTP_REQUEST_METHOD] = normMethod;
+    attributes[ATTR_URL_FULL] = url;
```
overbalance: Nav instrumentation uses `defaultSanitizeUrl` here. Should this match?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+   * Finish span, add attributes, network events etc.
+   * @param span
+   * @param response
+   */
+  private _endSpan(span: Span, response: FetchResponse) {
+    span.setAttribute(ATTR_HTTP_RESPONSE_STATUS_CODE, response.status);
+    // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status
+    if (response.status >= 400) {
+      span.setStatus({ code: SpanStatusCode.ERROR });
+      span.setAttribute(ATTR_ERROR_TYPE, String(response.status));
```
overbalance: On network errors this sets `status_code = 0` and `error.type = "0"`. Semconv says to skip the status code when there is no response, and set `error.type` to the error name (like `"TypeError"`) instead of `"0"`.

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+              error,
+            );
+            instrumentation._endSpan(span, {
+              status: 0,
+            });
+          }
+          return response;
+        }
+
+        function onError(span: Span, error: FetchError): never {
```
overbalance: A rejected fetch can reject with anything (`TypeError`, `DOMException`, etc.), not always a `FetchError`. The type here (and around 207) says more than we really know. It won't crash thanks to the `|| 0` and `error.message` fallbacks, but typing the reason as `unknown` and narrowing with a small helper would be more correct.

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+    DIAG_LOGGER.warn('Platform has ReadableStream but not pipeThrough!');
+    return {
+      body,
+      length: Promise.resolve(undefined),
+    };
+  }
+
+  let length = 0;
+  let resolveLength: (l: number) => void;
+  const lengthPromise = new Promise<number>((resolve) => {
```
overbalance: This promise only resolves in `flush()`, which runs when the stream finishes normally. If the request is aborted or the body is never fully read, it never resolves, so the `.then`/`.catch` at `instrumentation.ts:149`/`157` never run and the promise stays pending. Resolving with `undefined` on cancel/error, plus a try/catch around `transform`, would fix it.

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+    resolveLength = resolve;
+  });
+
+  const transform = new TransformStream({
```
overbalance: `new TransformStream({...})` with no types is `TransformStream<any, any>`, so `chunk` is `any` and needs the `as Uint8Array` cast on line 102. `new TransformStream<Uint8Array, Uint8Array>(...)` removes both the cast and the implicit `any`.
```suggestion
  const transform = new TransformStream<Uint8Array, Uint8Array>({
```

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+ *   in which case the second arg has type RequestInit
+ *   RequestInit is NOT cloneable, but RequestInit.body is writable
+ *   so we can chain it into ReadableStream.pipeThrough()
+ *
+ *   ReadableStream.pipeThrough() lets us process a stream and returns a new stream
+ *   So we can measure the body length as it passes through the pie, but need to attach
+ *   the new stream to the original request
+ *   so that the browser still has access to the body.
+ *
+ * @param body
```
overbalance: `@param body` does not match the signature (`...args: Parameters<typeof fetch>`)

packages/instrumentation/src/utils/url.ts
```diff
@@ -54,3 +54,53 @@ export function defaultSanitizeUrl(url: string): string {
     return sanitized;
   }
 }
+
+/**
+ * Parses url using URL constructor or fallback to anchor element.
```
overbalance:
```suggestion
 * Parses a URL, resolving relative URLs against document.baseURI (or location.href)
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+    }
+
+    return this.tracer.startSpan(name, {
+      kind: SpanKind.CLIENT,
+      attributes,
+    });
+  }
+
+  /**
+   * Finish span, add attributes, network events etc.
```
overbalance: `_endSpan` does not add network events, it only sets the status code and ends the span
```suggestion
   * Finish span, add attributes, etc.
```

packages/instrumentation/package.json
```diff
@@ -61,7 +61,9 @@
     "@opentelemetry/api": "^1.9.1"
   },
   "devDependencies": {
-    "@opentelemetry/sdk-logs": "0.219.0"
+    "@opentelemetry/sdk-logs": "0.219.0",
+    "@opentelemetry/sdk-trace-base": "^2.7.1",
+    "msw": "^2.14.6"
```
overbalance:
```suggestion
    "@opentelemetry/sdk-trace-base": "2.7.1",
    "msw": "2.14.6"
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+
+const hasBrowserPerformanceAPI = typeof PerformanceObserver !== 'undefined';
+
+export class FetchInstrumentation extends InstrumentationBase<FetchInstrumentationConfig> {
+  // Note: Intentionally *not* using `_enabled` as the field name to avoid
+  // any possible confusion with the `_enabled` field used on the *Node.js*
+  // InstrumentationBase class.
+  // Also not initializing the fields to `false` because the base class
+  // constructor already call `enable` modifying their values and it will
+  // set the instrumentaitons in a base state (enabled, patched but with flags set to false)
```
overbalance:
```suggestion
  // set the instrumentations in a base state (enabled, patched but with flags set to false)
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+  private declare _isEnabled: boolean;
+  private declare _isFetchPatched: boolean;
+
+  // To keep track of the resources for posterior cleanup the context registrey
```
overbalance:
```suggestion
  // To keep track of the resources for posterior cleanup the context registry
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+        }
+
+        function onSuccess(span: Span, response: Response): Response {
+          try {
+            // Clone the response and eagerly consume the clone to detect
+            // when the body transfer completes.  The original response is
+            // returned to the caller untouched so that it passes internal
+            // brand-checks required by APIs such as
+            // WebAssembly.compileStreaming.
+            // It consumens the entire body even if the user cancels reading it
```
overbalance:
```suggestion
            // It consumes the entire body even if the user cancels reading it
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+            // from the original response. But it does work with `AbortController.abort()`
+            // because it aborts the underliying fetch cancelling the original and clone streams
```
overbalance:
```suggestion
            // because it aborts the underlying fetch cancelling the original and clone streams
```

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,437 @@
...
+              } as PerformanceResourceTiming;
+              instrumentation._registerResource(createdSpan, resource);
+            });
+        });
+      };
+    };
+  }
+
+  /**
+   * Registers a resource and sets a timer for clearing the registry after a time bing idle
```
overbalance:
```suggestion
   * Registers a resource and sets a timer for clearing the registry after a time being idle
```

packages/instrumentation/src/utils/url.ts
```diff
@@ -54,3 +54,53 @@ export function defaultSanitizeUrl(url: string): string {
...
+  }
+  return false;
+}
+
+const HTTP_PORT_FROM_PROTOCOL: { [key: string]: string } = {
+  'https:': '443',
+  'http:': '80',
+};
+/**
+ * Extracts the server port fromt the given URL object
```
overbalance:
```suggestion
 * Extracts the server port from the given URL object
```

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+  let length = 0;
+  let resolveLength: (l: number) => void;
+  const lengthPromise = new Promise<number>((resolve) => {
+    resolveLength = resolve;
+  });
+
+  const transform = new TransformStream({
+    start() {},
+    async transform(chunk, controller) {
+      const bytearray = (await chunk) as Uint8Array;
```
overbalance:
```suggestion
      const bytearray = await chunk;
```

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+ *   However, reading the body here can only be done async; the body() method returns a promise
+ *   this means this entire function has to return a promise
+ *
+ * OR the first arg is a url/string
+ *   in which case the second arg has type RequestInit
+ *   RequestInit is NOT cloneable, but RequestInit.body is writable
+ *   so we can chain it into ReadableStream.pipeThrough()
+ *
+ *   ReadableStream.pipeThrough() lets us process a stream and returns a new stream
+ *   So we can measure the body length as it passes through the pie, but need to attach
```
overbalance:
```suggestion
 *   So we can measure the body length as it passes through the pipe, but need to attach
```

packages/instrumentation/src/fetch/instrumentation.ts
joaquin-diaz: My original idea was to call `unregister` in the resource-timing instrumentation once the context is consumed given that for network calls there's a 1:1 relationship between span context and resource-timing entries

packages/instrumentation/src/resource-timing/instrumentation.ts
```diff
@@ -272,7 +268,13 @@ export class ResourceTimingInstrumentation extends InstrumentationBase<ResourceT
...
           [ATTR_RESOURCE_RENDER_BLOCKING_STATUS]: entry.renderBlockingStatus,
         },
-      });
+      };
+      const ctx = getNetworkContextRegistry().getContext(entry);
+
+      if (ctx) {
+        record.context = ctx;
+      }
+      this.logger.emit(record);
```
joaquin-diaz, line 279: I think it's safe to call unregister here given that this context wouldn't be used in another entry?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,430 @@
...
+        err,
+      );
+    }
+  }
+
+  override disable(): void {
+    if (!hasBrowserPerformanceAPI) {
+      return;
+    }
+    if (!this._isEnabled) {
```
joaquin-diaz, line 102: nit: can we be consistent with spaces here? In enable you're leaving a white space between if statements, but not here

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,430 @@
...
+    });
+  }
+
+  /**
+   * Finish span, add attributes, network events etc.
+   */
+  private _endSpan(span: Span, response: FetchResponse) {
+    span.setAttribute(ATTR_HTTP_RESPONSE_STATUS_CODE, response.status);
+    // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status
+    if (response.status >= 400) {
```
joaquin-diaz: in `endSpanOnError` you use `0` as a fallback when the status is not present (or it's 0)
```
instrumentation._endSpan(span, {
  status: error.status || 0,
  statusText: error.message,
});
```
That means that `span.setStatus({ code: SpanStatusCode.ERROR })` wouldn't be set for 0

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Helper function to determine payload content length for XHR requests
+ * @param body
+ * @returns content length
+ */
+export function getXHRBodyLength(
+  body: Document | XMLHttpRequestBodyInit,
+): number | undefined {
+  if (isDocument(body)) {
+    return new XMLSerializer().serializeToString(document).length;
```
joaquin-diaz: Shouldn't you serialize `body` instead of `document`?

packages/instrumentation/src/resource-timing/instrumentation.test.ts
```diff
@@ -583,6 +587,48 @@ describe('ResourceTimingInstrumentation', () => {
...
+      observerCallback(
+        createMockPerformanceObserverEntryList([mockEntry]),
+        mockObserver as unknown as PerformanceObserver,
+      );
+
+      triggerIdleCallback(10);
+
+      const records = inMemoryExporter.getFinishedLogRecords();
+      expect(records).toHaveLength(1);
+      console.log(records[0]);
```
joaquin-diaz: Remove debug line

packages/instrumentation/src/utils/request.ts
joaquin-diaz: We don't need await once we add the type in TransformStream

packages/instrumentation/src/fetch/instrumentation.test.ts
```diff
@@ -0,0 +1,656 @@
...
+
+    it('should not unwrap global fetch when disabled', () => {
+      expect(isWrapped(globalThis.fetch)).toBeFalsy();
+      instrumentation = new FetchInstrumentation();
+      expect(isWrapped(globalThis.fetch)).toBeTruthy();
+      instrumentation.disable();
+      expect(isWrapped(globalThis.fetch)).toBeTruthy();
+    });
+
+    it('should not wrap global fetch when instantiated with `enabled: false`', () => {
```
joaquin-diaz: Is this the same test as line 203?

packages/instrumentation/src/fetch/instrumentation.test.ts
```diff
@@ -0,0 +1,656 @@
...
+    // related to CORS requests.
+    describe('cross origin requests', () => {
+      const corsFetch = () =>
+        fetch('http://example.com/api/status.json', {
+          mode: 'cors',
+          headers: { 'x-custom': 'custom value' },
+        });
+      it.skip('should create a span with correct root span', async () => {
+        await corsFetch();
+        // XXX: implement tests
```
joaquin-diaz: In the description is says:
```
// So the purpose of this test module is mostly just to test the configs
// related to CORS requests.
```

But you're skipping the test. Is the description wrong or are we missing a test?

joaquin-diaz:
> @joaquin-diaz
> 
> I've updated this PR to use the `NetworkContextRegistry` from #329. Now `fetch` and `resource-timing` use the registry to correlate the logs to the fetch spans.
> 
> Now is good for review but I was thinking to include the e2e tests you implemented in #283. WDYT?

Makes sense! https://github.com/open-telemetry/opentelemetry-browser/pull/332

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: I haven't thought about polling, thanks. The idea was to keep the resource for a while so it can be accessed for more instrumentations if necessary.

I can move the logic there and unregister in resource-timing since for now there isn't any other instrumentation that needs the context.

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: fetch instrumentation in core does not but I think it should.

packages/instrumentation/src/utils/request.ts
david-luna: right, also I think is redundant to add param names in the comments

david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/d7803db9ebaabc3a7a8008666a0fa92036a9cba1

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/301bb0413c07e51dbfd6355a0ac1d4735859ec4c

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: > What do you think about registering from _endSpan instead, when reading is really done?

I think is a good idea :)

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: moved the `unregister` to resource timing instrumentation. the unregistration happens just after correlating the log with the registered context. commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/e802204b55b88d0650a8993ad7688f9e1d3a9e8b

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/e802204b55b88d0650a8993ad7688f9e1d3a9e8b

note: for some test I had to consume the body to get the right timing assertions

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/9a089777601d71d190efe1781e651fe4258d56b6

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: I'm going to apply your suggestion although I'm not sure what will happen regarding timings. do you think its possible in this situation that the time here would be lower than the one reported in resource timings?

david-luna: https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/638744b6eca228944d10809ee1fdf0a72771ca86

packages/instrumentation/src/fetch/instrumentation.test.ts
david-luna: indeed it is. Will remove in a follow up commit

packages/instrumentation/src/fetch/instrumentation.test.ts
david-luna: We're missing them. Will ad them shortly

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/53e4d68637148000c18d4e10787d8a11b00dcfad

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/53e4d68637148000c18d4e10787d8a11b00dcfad

packages/instrumentation/src/resource-timing/instrumentation.ts
david-luna: yes! done in the `if(ctx)` block above so we only call unregister if it has some effect.

packages/instrumentation/src/utils/request.ts
david-luna: IIUC then I need to wrap the `ReadableStream` within another one to be able to set the `cancel` handler and:
- resolve to `undefined`?
- resolve to the current value of lenght?

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/5f19ac608fcc5b8d209e67af8777b673b53657be

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: also commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/9bbcf079c6d2fdbf2aeb14240365b208288a3263

packages/instrumentation/src/resource-timing/instrumentation.test.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/9a438f28108daa55e539f5c578775f7369417c52

packages/instrumentation/src/utils/request.ts
david-luna: this comes from the implementation in https://github.com/open-telemetry/opentelemetry-js/blob/636e27aa033c8f75c1a1bbae89371da11465edda/experimental/packages/opentelemetry-instrumentation-fetch/src/utils.ts#L124

not sure if it was intentional (the param is the actual document) or not. I'll make the change and add a test

david-luna: since the function will be used for XHR instrumentation I'll add the test in a follow up PR with the XHR instrumentation tests.

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -0,0 +1,425 @@
...
+        span.setAttribute(ATTR_ERROR_TYPE, response.statusText);
+      }
+    }
+    span.end();
+
+    // Register the resource context for other instrumentations
+    const resource = this._spanResources.get(span);
+    if (resource) {
+      this._spanResources.delete(span);
+      getNetworkContextRegistry().register(span, {
```
mquentin, line 338: I think we might a risk of overgrowing Registry. From What I understand (sorry if I missed something) the `NetworkContextRegistry` is `unregister` only when there is a successful `_emitResource` but if some resources are ignored by 

```
const shouldIgnoreUrl = matchesUrl(entry.name, this._config.ignoreUrls);
```
It will not be `unregister` isn't it ?


Suggestion:

I feel it's hard to make sure 100% that the `unregister` will be ever call so I suggest we add a small check in `packages/instrumentation/src/utils/NetworkContextRegistry.ts` where, when registering a new entry, if we've reached a limit already, we unregister first the oldest entry as a preventive step.

mquentin (commented):
Small suggestion about `NetworkContextRegistry` management

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: I think you're right. There are situations where `_emitResource` from resource-timing instrumentation might not be called:
- url is ignored in that instrumentation
- instrumentation is disabled

I remember the 1st approach was for fetch instrumentation to unregister after a TTL but it was discarded since handling timers seemed to convoluted. 

>Suggestion:
>
>I feel it's hard to make sure 100% that the unregister will be ever call so I suggest we add a small check in packages/instrumentation/src/utils/NetworkContextRegistry.ts where, when registering a new entry, if we've reached a limit already, we unregister first the oldest entry as a preventive step.

I think is a good idea but the implementation uses arrays when different entries collide with the same key. I think this needs a refactor on the `ContextRegistry` classes. I wonder if it makes more sense to have a separate PR for this and merge it before this PR.

mquentin: Indeed, having a separate PR make sens to me as this fetch instrumentation is only a consumer of the ContextRegistry class. I am also convinced that the registry management safeguards should be 100% handled by the ContextRegistry itself to prevent risk of overgrowing Registry.

Also this PR is already pretty big and merging it would unlock the package migration so I will approve the PR

mquentin (approved):
Taking in consideration the followup PR https://github.com/open-telemetry/opentelemetry-browser/pull/281#discussion_r3673248236 and to unlock the package migration in the meantime, LGTM

packages/instrumentation/src/fetch/instrumentation.ts
joaquin-diaz: Originally I didn't want to include any TTL logic in the base `ContextRegistry` since it's hard to make a rule that would apply to all consumers, but I think it would be ok to define per-domain rules like in `NetworkContextRegistry`. 

Also I would imagine if a URL is ignored, then it should also be ignored for all instrumentations, that's something we might need to enforce through the SDK

joaquin-diaz (approved):

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: https://github.com/open-telemetry/opentelemetry-browser/pull/368 is merged and this PR has been synced with main. This should not be a problem anymore :)

packages/instrumentation/src/utils/request.ts
david-luna: @overbalance there is an alternative to calculate the size if we use [ReadableStream.tee()](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee). We could return one stream for the app and use the other to calculate the size.

The only con I see is we read the entire body separately regardless if the reads the body or not. Are we interested to know the body size even if the application hasn't consumed it all?

The former fetch instrumentation has the same logic 
https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-fetch/src/utils.ts#L92-L108

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: pr merged and synced with this one

packages/instrumentation/src/utils/request.ts
david-luna: Wrapping in another readable stream was simple enough. Commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/7eb12b8d3cc625e17901239792b1a13da9ba24ea

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,200 @@
...
+      controller.enqueue(value);
+    },
+    cancel(reason) {
+      resolveLength(undefined);
+      bodyReader.cancel(reason);
+    },
+  });
+
+  return {
+    // body: body.pipeThrough(transform),
```
joaquin-diaz: Remove this if is not necessary

packages/instrumentation/src/utils/request.ts
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/* eslint-disable baseline-js/use-baseline */
```
joaquin-diaz: Can you only disable the lines we need to be disabled? I'm worried that we might miss something in the future having this disabled for the entire file

packages/instrumentation/README.md
joaquin-diaz, line 18: Should we add some documentation to describe the connection between resource timing and fetch when both are used?

packages/instrumentation/src/utils/request.ts
david-luna: this is a leftover. Thanks for catching it. I've disabled only where needed.
commit https://github.com/open-telemetry/opentelemetry-browser/pull/281/commits/644fe478c0031c87d58fee9ca4b47ed1934d9fc8

packages/instrumentation/src/utils/request.ts
david-luna: done!

packages/instrumentation/README.md
david-luna: I think is a good idea :)

We could add a note in each instrumentation referring the other one or add a specific section. I'm biased towards the later and I can add such section in the [follow-up PR](https://github.com/open-telemetry/opentelemetry-browser/pull/367) I have for XHR.

PR #283, opened 2026-05-15
feat: migrate fetch instrumentation from opentelemetry-js

joaquin-diaz:
## Summary

This ports `@opentelemetry/instrumentation-fetch` from [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-fetch) into this browser-only repo, exposed as `@opentelemetry/browser-instrumentation/experimental/fetch`.

### Behavioral

- **Stable semconv only.** Old attributes (`http.status_code`, `http.host`, `http.url`, `http.method`, `http.scheme`, `http.user_agent`, `component`, etc.) and the `semconvStabilityOptIn` config option have been removed. Only the stable set is emitted: `http.request.method`, `url.full`, `http.response.status_code`, `server.address`, `server.port`, `error.type`.

- **Network timing emitted as a log record instead of span events.** Upstream attaches one span event per timing field (`fetchStart`, `domainLookupStart`, `connectStart`, …) to the span. Here a single `LogRecord` is emitted with all timing values as attributes, correlated to the span via OTel context. The log carries an `eventName` (`browser.fetch.resource_timings` / `browser.fetch.cors_preflight_resource_timings`) and is skipped when `resource.startTime === 0`.

### Structural (no behavior difference in a browser)

- **Browser-only.** Node.js guards (`hasBrowserPerformanceAPI`, `typeof PerformanceObserver !== 'function'`) removed — this package never runs outside a browser.

- **No `@opentelemetry/sdk-trace-web` dependency.** `parseUrl`, `shouldPropagateTraceHeaders`, and `getResource` are ported locally under `src/utils/`. Logic is identical.

- **No `@opentelemetry/core` dependency.** `isUrlIgnored` and `millisToHrTime` are ported locally under `src/utils/`. Logic is identical.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ dist
 
 # AI agent rules
 AGENTS.md
+.claude
+
+# Test output
+.vitest-attachments
+**/*/coverage
--- a/e2e-tests/package.json
+++ b/e2e-tests/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "e2e-tests",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@opentelemetry/api": "^1.9.1",
+    "@opentelemetry/api-logs": "^0.217.0",
+    "@opentelemetry/browser-instrumentation": "file:../packages/instrumentation",
+    "@opentelemetry/exporter-logs-otlp-http": "^0.217.0",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.217.0",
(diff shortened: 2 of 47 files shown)


martinkuba:
This is related to https://github.com/open-telemetry/opentelemetry-browser/issues/259

joaquin-diaz:
Migrating in https://github.com/open-telemetry/opentelemetry-browser/pull/281

PR #288, opened 2026-05-22
feat: add `@opentelemetry/browser-sdk` package

david-luna:

Follow up on https://github.com/open-telemetry/opentelemetry-browser/pull/224. This PR is now targeting a feature branch so we can break down the implementation in smaller PRs and merge it to main once its ready.

- add a new `@opentelemetry/browser` package
- all comment in the prior PR were resolved except https://github.com/open-telemetry/opentelemetry-browser/pull/224#issuecomment-4491556702
- for this last comment the package now exports `startBrowserSdk` function instead of the combine utility. The reason is there is no other use case. Either you start one single SDK or both. The combine function stays internal until we decide to add a third signal (metrics).

This PR aims to define the main API to configure the signals (logs, traces) separate and together. Also it explores on how to provide a flexible way for defaults without having to pay the toll on bundle size.

## For those who want to read

Add a new package named `@opentelemetry/browser` which holds the setup of the different signals separately and provides a main function which composes them together into an SDK interface with `stop` method.

This approach has taken into consideration that configuring all signals into a single SDK class presents challenges when it comes to tree shaking and bundlers cannot anticipate which modules are going to be used. So the configuration has been split by signal and needs to be imported explicitly by the consumer making easier for bundlers to analyse the code and remove unused modules. The package provides different entry points depending on which signal you want to configure

| Subpath                                       | Purpose                                                 |
| ------------------------------- | ------------------------------------ |
| `@opentelemetry/browser/logs`    | `startLogsSdk` — setup logs         |
| `@opentelemetry/browser/traces` | `startTracesSdk` — setup traces |
| `@opentelemetry/browser/sdk`     | `startBrowserSdk` — setup both  |

with this approach you can configure a single signal

```js
import { startLogsSdk } from '@opentelemetry/browser/logs';

const logsSdk = startLogsSdk({
  logLevel: 'DEBUG',
  serviceName: 'my-service',
  exportConfig: {
    url: 'https://collector.mycompany.com/v1/logs',
    headers: { foo: 'bar' }
  },
  logRecordLimits: { attributeCountLimit: 128 },
});
```

or both at the same time

```js
import { startBrowserSdk } from '@opentelemetry/browser/sdk';

const sdk = startBrowserSdk({
  logLevel: 'DEBUG',
  serviceName: 'my-service',
  serviceVersion: '1.0',
  resourceAttributes: {
    'browser.has_speech_recognition': typeof SpeechRecognition === 'function'
  },
  exportConfig: {
    url: 'https://collector.mycompany.com',
    headers: { foo: 'bar' }
  },
  // Optional - logs signal configuration. Default empty empty
  logs: {
    logRecordLimits: { attributeCountLimit: 128 },
  },
  // Optional - traces signal configuration. Default empty empty
  traces: {
    spanLimits: { attributeCountLimit: 128 },
  },
});

```

**Note:** the intent is to have all necessary components to start/build SDKs in this package. This PR does not go any further than providing a small API (3 functions) and configurations types for several reasons:

- keep the review scope small so we have meaningful conversations. We can add more components in follow-up PRs also keeping the scope small
- this also enables other contributors to add code to this package so it's not a solo quest but rather a collaboration effort
- IMO as a consequence of the points above the velocity of implementing this package will increase. Also I will not be holding/blocking anything in a huge PR (this is already big enough).

Also this PR targets a feature branch and not main so the package won't be published until the SIG decides it has all it need to be public.

For more details on the package and its design refer to the `README.md` file of the package.

## Type of change

## How Has This Been Tested?

- [X] Unit tests

## Checklist:

- [X] Documentation has been updated


david-luna:
converted to draft in order to keep the `squash and merge` button disabled

examples/with-tracing/src/main.ts
```diff
@@ -60,11 +60,11 @@ registerInstrumentations({
 });
 
 // --- Button handlers ---
-document.getElementById('fetch-button')!.addEventListener('click', () => {
+document.getElementById('fetch-button')?.addEventListener('click', () => {
```
david-luna, line 63: note to reviewer: I think this is the result of linting. Is it possible the rules changed?

.github/workflows/scorecard.yml
```diff
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@9e0d7b8d25671d64c341c19c0152d693099fb5ba # v4.35.5
+        uses: github/codeql-action/upload-sarif@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
```
david-luna: note for reviewer: sync form the main branch

packages/sdk/src/context.ts
```diff
@@ -0,0 +1,67 @@
+/*
```
david-luna, line 1: note for reviewer: this is not exported yet. I hope it serves to start a conversation about defaults that are components (contextManager, propagators, etc...).

IMO if the SDK functions resolves defaults for context manager, propagators and others it will make difficult for bundlers to tree shake if not used. The proposal is to expose functions that provide the defaults but user has to explicitly set it. Something like.

```js
import { startBrowserSdk } from '@opentlemetry/browser/sdk';
// import subpath so it easier to tree shake
import { getDefaultContextManager } from '@opentlemetry/browser/context';

const sdk = startBrowserSdk({
  traces: {
    contextManager: getDefaultContextManager(),
  },
});
```

not using the import explicitly takes out the `context` module from the graph so it's not included in the final bundle.

packages/sdk/src/traces.ts
```diff
@@ -0,0 +1,85 @@
...
+          url: tracesEndpoint,
+          headers: config?.exportConfig?.headers,
+        }),
+        config?.processorConfig,
+      ),
+    );
+  }
+
+  const tracerProvider = new BasicTracerProvider({
+    // sampler: new TraceIdRatioBasedSampler(
```
david-luna, line 81: following up https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes#r3310989075

If we set a default sampler here it means its code will be in the bundle regardless if being used or not. that's why the approach explained in the previous comment is proposed

packages/browser/src/diag.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+import { DiagConsoleLogger, DiagLogLevel, diag } from '@opentelemetry/api';
+
+/**
+ * Sets the global logger with the given log level. If called a second time
+ * the operation will be a noop
+ */
+let loggerSet = false;
+export function setSdkLogger(level: keyof typeof DiagLogLevel) {
```
joaquin-diaz: I like the runtime check just in case. Also, I would probably add a default and make the argument optional

packages/sdk/src/traces.ts
```diff
@@ -0,0 +1,85 @@
...
+    resourceFromAttributes(resourceAttributes),
+  );
+
+  // Resolve the list of span processors.
+  // - if provided by the user use them
+  // - otherwise create a `BatchSpanProcessor`
+  const spanProcessors: SpanProcessor[] = config?.processors || [];
+
+  if (config?.processors) {
+    spanProcessors.push(...config.processors);
```
joaquin-diaz, line 52: Wouldn't this push the processors twice?

```
// Here if config?.processors exists then spanProcessors = config.processors
const spanProcessors: SpanProcessor[] = config?.processors || [];

// if config?.processors exists pushes config.processors to spanProcessors
if (config?.processors) {
    spanProcessors.push(...config.processors);
}

File: packages/sdk/src/types.ts [hunk 7]
```diff
@@ -0,0 +1,163 @@
...
+   * Configuration for the LogRecord exporter.
+   */
+  exportConfig?: ExportConfig;
+  /**
+   * Limits for each LogRecord.
+   */
+  logRecordLimits?: LogRecordLimits;
+  /**
+   * List of LogRecordProcessor for the logger provider. Setting this will make the SDK
+   * ignore `processorConfig` and `exportConfig` since no `BatchLogRecordProcessor` will
```
joaquin-diaz: Given that probably `BatchLogRecordProcessor` and `BatchSpanProcessor` would be the default processors for exporting telemetry in most cases, I am not sure if maybe is worth always using that as a default even if the user wants to use their own processors. As a user if I want to use a custom processor then I would also need to configure BatchLogRecordProcessor/BatchSpanProcessor adding complexity to my setup. Maybe as part of `exportConfig` we can take a "export processor" allowing to customize or something similar

packages/browser/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+The following configuration options are common to each independent SDK. When usgin `startBrowserSdk` which
+combines all signals these options are accepted and the top level of the configuration and they don't need to
+be in signal specific configuration. See the example below.
+
+```javascript
+// Signal SDK accepts some configs
+startLogsSdk({ serviceName: 'my-serivce' });
+
+// When startnig all at once...
```
joaquin-diaz:
```suggestion
// When starting all at once...
```

packages/browser/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### resourceAttributes
+
+Key-value pairs to be used as resource attributes.
+
+### Logs configuration
+
+#### processorConfig
+
+Object containing configuraiton options for the batch processing of log records. These options are:
```
joaquin-diaz:
```suggestion
Object containing configuration options for the batch processing of log records. These options are:
```

The typo is repeated a few more times below

packages/sdk/src/types.ts
joaquin-diaz, line 1: Given that this would serve as documentation when writing code:

1. Some configuration items have missing descriptions, should we add it to all of them?
2. Should we show default values? For example:
```
/**
  Max number of signals per batch. 
  @default 12345
*/
maxExportBatchSize?: number;
```

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,153 @@
...
+  WebSdkFactory,
+} from './types.ts';
+
+interface SdkFactories {
+  logs?: WebSdkFactory<LogsConfig>;
+  traces?: WebSdkFactory<TracesConfig>;
+}
+
+/**
+ * Utility funcitons to extract the configurations from the factory
```
joaquin-diaz:
```suggestion
 * Utility functions to extract the configurations from the factory
```

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,153 @@
...
+    rootConfig.resourceAttributes ??= {};
+    if (rootConfig.serviceName) {
+      rootConfig.resourceAttributes['service.name'] = rootConfig.serviceName;
+    }
+    if (rootConfig.serviceVersion) {
+      rootConfig.resourceAttributes['service.name'] = rootConfig.serviceVersion;
+    }
+
+    // Export
+    rootConfig.exportConfig = Object.assign(
```
joaquin-diaz: Any reason to not use object spreading?

packages/browser/src/traces.test.ts
```diff
@@ -0,0 +1,213 @@
...
+
+import { startTracesSdk } from './traces.ts';
+import type { WebSdk } from './types.ts';
+
+const BSP_SCHEDULE_DELAY = 10;
+
+describe('startTracesSdk', () => {
+  const response = { ok: true, json: async () => ({ ok: true }) } as Response;
+  const fetchSpy = vi.spyOn(globalThis, 'fetch').mockResolvedValue(response);
+  const setGlobalTracerProviderSpy = vi
```
joaquin-diaz: Do we need these spies? In instrumentation tests we allow the sdk to work as intended and then we use memory exporters to validate output. I guess the fetch spy is the only way of validating we set the right URL to the exporter, in that case I would have a test testing that and using the memory exporter in the rest

packages/browser/src/traces.test.ts
```diff
@@ -0,0 +1,213 @@
...
+  });
+
+  it('should register a TracerProvider with a BatchSpanProcessor', async () => {
+    // Act
+    tracesSdk = startTracesSdk();
+
+    // Assert
+    expect(setGlobalTracerProviderSpy).callCount(1);
+    // biome-ignore lint/suspicious/noExplicitAny: accessing private props
+    const processors = (trace.getTracerProvider() as any)[
```
joaquin-diaz: Instead of relying on internal methods and mocking we could set up the memory exported and if we get spans then it means the batch processor is working. I'd try to avoid having tests using internal methods / attributes, I see you do the same on the other tests

joaquin-diaz (commented):
Added a couple of comments but I like how this is going!

packages/browser/src/sdk.ts
david-luna: old habit to support certain browsers and node versions. will change it in a follow up commit

packages/browser/src/traces.test.ts
david-luna: If I'm not mistaken the API only allows to register a provider (logs, traces) once here https://github.com/open-telemetry/opentelemetry-js/blob/c0bfea73a2304eaf055403c7858aa74300cb2e1e/api/src/api/trace.ts#L58 (it does not user the `allowOverride` param of the register function)

So only the 1st call to `startTraceSdk` would actually register a provider. Ignoring any configuration from the other tests.

packages/browser/src/traces.test.ts
david-luna: I like the idea :)

packages/sdk/src/traces.ts
david-luna: yes. `const spanProcessors: SpanProcessor[] = config?.processors || [];` is a leftover of a previous idea which I discarded lated. Good catch

packages/sdk/src/types.ts
david-luna: I guess the default values would not change much over time. So this can be done withotut adding extra maintanance

packages/sdk/src/types.ts
david-luna: I thought about several combinations like:
- create `BatchSpanProcessor` and `BatchLogRecordProcessor` if the `exportConfig` is set. In most (all?) cases users don't want to send data to "http://localhost:4318"
- allow an `exporters` config option per signal so the batch processors will use them

I'm inclined towards the 1st but there is the rare case a user uses custom processors and wants to export data to localhost (Electron apps maybe?). If we go that path I think we should be clear in the documentation and `diag.debug` clearly in this case.

packages/sdk/src/types.ts
david-luna: 🤔 the start functions do not set these defaults. They are set in the underlying components (providers, etc.)
- should we set it here to avoid surprises when we update dependencies?
- or leave it as is?

packages/browser/src/traces.test.ts
david-luna: Following up on this. I think the next test already covers that since we catch the fetch request and assert on the export. I think we could use the `InMemorySpanExporter` to test the `processors` option once we decide what to do about https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes#r3318291251

packages/sdk/src/traces.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes/fb9df6cfe0ebe426e10c471be3d5bfef097630ca

packages/sdk/src/types.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes/fb9df6cfe0ebe426e10c471be3d5bfef097630ca

packages/browser/src/diag.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes/1880cff7a6b7ca46bf70ff2302ff05d6414925ef

martinkuba:
I wonder if the directory this package is in should be called `sdk` instead of `browser`, which does not describe what the package is (even though it maps to the published name, but that has a specific reasoning). We already don't match it, as we have `instrumentation` directory that contains package called `browser-instrumentation`.

packages/sdk/src/types.ts
joaquin-diaz: First option makes sense. As long as there's a way to customize whatever default setting we have I think we should try to make all the default settings as easy as possible for the first time user

packages/sdk/src/types.ts
joaquin-diaz: > should we set it here to avoid surprises when we update dependencies?

That actually happened to us hah. Given that we want to detach as much as possible from the js repos maybe we want to set defaults here?

packages/browser/src/traces.test.ts
joaquin-diaz: If we call `trace.disable()` and `logs.disable()` after each test then I think that removes it?

packages/browser/src/traces.test.ts
joaquin-diaz: Right, you wouldn't get any request if the batch processor isn't working

packages/sdk/src/types.ts
david-luna: fine for me :)

packages/browser/src/traces.test.ts
david-luna: That might work. I'll try and push a new commit if so

david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/commits/ab0e5ad6915b9e418367c9a75ba28848f17d8923

packages/browser/src/traces.test.ts
david-luna: as for now I've removed the tests that inspect internal properties in https://github.com/open-telemetry/opentelemetry-browser/pull/288/commits/ab0e5ad6915b9e418367c9a75ba28848f17d8923

david-luna:
> I wonder if the directory this package is in should be called `sdk` instead of `browser`, which does not describe what the package is (even though it maps to the published name, but that has a specific reasoning). We already don't match it, as we have `instrumentation` directory that contains package called `browser-instrumentation`.

If we want to align with the existing instrumentation package then we should:
- change the package name to `@opentelemetry/browser-sdk`
- rename the folder to `sdk`

I'm okay with that. I would add a review to the package exports section for better naming. This might look weird

```js
import { strartBrowserSdk } from '@opentelemetry/browser-sdk/sdk';

// rest of the code
```

packages/sdk/src/types.ts
david-luna: I've implemented the 1st option and also added a `quickStartBrowserSdk` function (probably will be renamed) which accepts only few config options to make it easier to start

packages/sdk/src/types.ts
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * @defaultValue undefined
+   */
+  headers?: Record<string, string>;
+}
+
+/**
+ * Batch processor configuration. Can be used globally or per signal
+ */
+export interface ProcessorConfig {
```
martinkuba: All of these are specifically batch processor configs. At first I thought it was generic processors configuration. I would suggest that we rename this to BatchProcessorConfig (or something similar) to make it clear what it represents.

Btw, at first I was going to suggest we fold ProcessorConfig into ExportConfig, since they are used together and ProcessorConfig basically only applies to (default) export. But that might be confusing as well.

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,143 @@
...
+    // - accept resource detectors?
+    // - how to avoid creating different resources (here and in signals)
+    //   - in the config? may be misleading for users seeing
+    //   - maybe using an internal module with set/get like diag
+    rootConfig.resourceAttributes ??= {};
+    if (rootConfig.serviceName) {
+      rootConfig.resourceAttributes['service.name'] = rootConfig.serviceName;
+    }
+    if (rootConfig.serviceVersion) {
+      rootConfig.resourceAttributes['service.name'] = rootConfig.serviceVersion;
```
martinkuba: This should be `service.version`

packages/sdk/src/types.ts
```diff
@@ -0,0 +1,217 @@
...
+ * these properties belong to the root configuration and not to
+ * the sginal specific config.
+ */
+export interface CommonConfig {
+  /**
+   * Set `disabled: true` to disable the SDK
+   *
+   * @defaultValue undefined
+   */
+  disabled?: boolean;
```
martinkuba, line 87: It does not look like this is ever used

packages/sdk/src/sdk.ts
```diff
@@ -0,0 +1,143 @@
...
+        tracesConfig.exportConfig.url = endpointUrl.href;
+      }
+      tracesConfig.resourceAttributes = rootConfig.resourceAttributes;
+      sdks.push(factories.traces(tracesConfig));
+    }
+
+    return {
+      shutdown() {
+        return Promise.allSettled(sdks.map((s) => s.shutdown())).then(
+          () => undefined,
```
martinkuba: This swallows any rejections, which is a different behavior than when using the logs/trace SDKs separately. I think we should match them, and if we decide to never reject, then I think we should at least log it.

packages/sdk/src/sdk.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    // Export
+    rootConfig.exportConfig = {
+      url: DEFAULT_OTLP_ENDOINT,
+      ...rootConfig.exportConfig,
+    };
+
+    const sdks: WebSdk[] = [];
+    const endpointUrl = new URL(
```
martinkuba: If this throws because of malformed url, should we log why?

packages/sdk/src/types.ts
```diff
@@ -0,0 +1,217 @@
...
+   *
+   * @defaultValue undefined
+   */
+  propagators?: TextMapPropagator[];
+  /**
+   * Sampler to be used by tracer to decide if a Span os sampled or not.
+   *
+   * @defaultValue undefined
+   */
+  sampler?: Sampler;
```
martinkuba, line 185: It looks like this is never used

packages/browser/src/sdk.ts
```diff
@@ -0,0 +1,143 @@
...
+ * available at the config root)
+ */
+type RemoveCommonProps<T> = Omit<T, keyof CommonConfig>;
+type ExtractConfigs<T> = Partial<{
+  [K in keyof T]: T[K] extends WebSdkFactory<infer C>
+    ? RemoveCommonProps<C>
+    : never;
+}>;
+
+const DEFAULT_OTLP_ENDOINT = 'http://localhost:4318';
```
martinkuba: typo

packages/browser/src/traces.ts
```diff
@@ -0,0 +1,86 @@
...
+} from '@opentelemetry/resources';
+import type { SpanProcessor } from '@opentelemetry/sdk-trace-base';
+import {
+  BasicTracerProvider,
+  BatchSpanProcessor,
+} from '@opentelemetry/sdk-trace-base';
+import { setSdkLogger } from './diag.ts';
+import type { TracesConfig, WebSdk } from './types.ts';
+
+const DEFAULT_TRACES_OTLP_ENDOINT = 'http://localhost:4318/v1/traces';
```
martinkuba: typo

martinkuba (changes requested):

packages/sdk/src/types.ts
david-luna: >  I would suggest that we rename this to BatchProcessorConfig (or something similar) to make it clear what it represents.

I think is a good idea :)

packages/browser/src/sdk.ts
david-luna: good catch

packages/sdk/src/types.ts
david-luna: I was following the nodejs SDK implementation. I think it is useful if the user wants to block SDK start in certain situations that can only be detected when the server is rendering the page or the HTML already in the browser

```js
function isBot() {
  // some logic to detect bots based on `navigator` or other APIs
}

startBrowserSdk({
  disabled: isBot(), // the user does not want data fro bots
  // rest of the confguration
})
```

packages/sdk/src/types.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/commits/d89c08233985b8b4309df723ecc4efc03c31bdce

packages/sdk/src/types.ts
david-luna: I've just added a `sampler.js` file in https://github.com/open-telemetry/opentelemetry-browser/pull/288/commits/d89c08233985b8b4309df723ecc4efc03c31bdce to show how it could be used. 

```js
import { AllwaysOnSampler } from '@opentelemetry/browser-sdk/sampler';
import { startTracesSdk } from '@opentelemetry/browser-sdk/traces'

startTracesSdk({
  sampler: new AllwaysOnSampler(),
  // ...
})
```

packages/browser/src/sdk.ts
david-luna: fixed!

packages/browser/src/traces.ts
david-luna: fixed

packages/sdk/src/types.ts
martinkuba: What I meant is that the SDK currently does not check the value of this config.

packages/sdk/src/sdk.ts
david-luna: I've added a check of the URL and an early return if it can't be parsed. Commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes/1ac1d6a80937de3fb7113d6b0f4eb444f528396c#diff-04689b788dad65b6ce4b43e7344d3b854b1a95ad2c8692c6ece28ccd7d9a96b4

packages/sdk/src/sdk.ts
david-luna: Now the shutdown fails joining all the possible failures. Commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/changes/1ac1d6a80937de3fb7113d6b0f4eb444f528396c#diff-04689b788dad65b6ce4b43e7344d3b854b1a95ad2c8692c6ece28ccd7d9a96b4

packages/sdk/src/types.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/288/commits/fa2d8a1002f3a41924a2ec1bd037557f8192786b

martinkuba (approved):

overbalance (approved):

packages/sdk/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "traces"
+  ],
+  "homepage": "https://github.com/open-telemetry/opentelemetry-browser",
+  "bugs": "https://github.com/open-telemetry/opentelemetry-browser/issues",
+  "license": "Apache-2.0",
+  "author": "OpenTelemetry Authors",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/open-telemetry/opentelemetry-browser.git",
+    "directory": "packages/browser"
```
overbalance, line 21: TODO: fix after merge

PR #300, opened 2026-06-04
chore(deps): update @arethetypeswrong/core to 0.18.3

overbalance:

Routine dependency maintenance. Bumps the `@arethetypeswrong/core` devDependency from `0.18.2` to `0.18.3`.

The repo carried an `overrides` entry pinning `fflate` to `0.8.2` for `@arethetypeswrong/core`. Release `0.18.3` declares `fflate@^0.8.3`, so the override is no longer needed and is removed. The lock file now resolves `fflate@0.8.3`.

- Update `@arethetypeswrong/core` `0.18.2` -> `0.18.3` in `package.json` and `package-lock.json`.
- Remove the now-redundant `overrides.@arethetypeswrong/core.fflate` pin.

## Type of change

## How Has This Been Tested?

- [x] Local install resolves cleanly and the workspace builds with the updated dev dependency.

## Checklist:


package-lock.json
```diff
@@ -958,422 +958,6 @@
         "tslib": "^2.4.0"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
```
trentm, line 961: Possibly you ran `npm install` with one of those buggy npm versions that drops platform-specific deps from package-lock?

Old thread: https://cloud-native.slack.com/archives/C08T7MZTV8W/p1761781178575799?thread_ts=1761762353.429559&cid=C08T7MZTV8W
https://github.com/npm/cli/pull/8431#issuecomment-3310130316

overbalance: ~~Other way around. "lockfile maintenance" keeps adding esbuild even though it's transitive, so any real `npm i` puts it right.~~

overbalance: nm, you were right. npm reverted to an older version after I reinstalled brew.

david-luna (approved):

PR #306, opened 2026-06-05
fix(dev): require npm 11.16.0

overbalance:

We can help protect this repo from the recent spate of npm supply chain attacks by requiring npm 11.16.0, which introduced [`npm approve-scripts`](https://docs.npmjs.com/cli/v11/commands/npm-approve-scripts):

> Manages the allowScripts field in your project's package.json, which records which of your dependencies are permitted to run install scripts (preinstall, install, postinstall, and prepare for non-registry sources). This command is the recommended way to maintain that field.

In npm 11.16.0 the `allowScripts` field is advisory. Install scripts still run, but npm warns on every install about any package whose scripts have not been reviewed, so a new dependency that gains an install script becomes immediately visible. A future npm release will block unreviewed install scripts, and adopting the field now means this repo gets that enforcement automatically. It is one layer, not a complete defense: it covers install-time scripts only, not malicious code in a package body that runs at import or build time.

- Require npm 11.16.0. `packageManager` and `devEngines.packageManager` are bumped from 11.9.0 to 11.16.0 with `onFail: "error"`, so older npm versions refuse to install.
- Review install scripts via `allowScripts`. Only `lefthook@2.1.8` (needed for git hook setup) is approved. Any other dependency running install scripts will produce an `npm warn allow-scripts` warning today and will be blocked once npm enforces the field.
- Update the Node runtime range to `^24 || ^26.3.0` and pin `.nvmrc` to 26.3.0 for development and CI.
- Update all six workflows to read the Node version from `.nvmrc` instead of `package.json`, since the runtime constraint is now a compound range that setup-node cannot resolve to a single version.
- Set `package-manager-cache: false` in all workflows. setup-node v6 enables the npm cache by default. As part of this hardening, CI prefers fresh installs from the registry over restored caches.

## Type of change

## How Has This Been Tested?

- `npm install` with an older npm (11.13.0) fails with a devEngines error as expected.
- `npm ci` with npm 11.16.0 succeeds, and the lefthook hook installation still runs via its `allowScripts` entry.
- Verified the advisory behavior with npm 11.16.0: a test dependency with an unapproved postinstall script still runs but triggers `npm warn allow-scripts` naming the package and the script.
- CI workflows on this PR run on Node 26.3.0 with caching disabled.

## Checklist:


joaquin-diaz (approved):

.github/workflows/release-please.yml
```diff
@@ -44,7 +44,6 @@ jobs:
       - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version-file: package.json
-          cache: "npm"
           registry-url: "https://registry.npmjs.org"
```
trentm, line 48: Perhaps go further and disable a package manager cache for this workflow, because it releases artifacts.
https://docs.zizmor.sh/audits/#cache-poisoning

This is done for setup-node via `with:\n  package-manager-cache: false`.
(I had done this for a couple workflows in opentelemetry-js recently, but missed "publish-to-npm.yml" which does the primary `npm publish` calls for that repo. face-palm.)

Though, zizmor doesn't flag this.
(I'm not exactly sure why. I'd have thought https://github.com/zizmorcore/zizmor/blob/8afebfe87b05ec2db1245a2f36300d3df2320add/crates/zizmor/src/audit/cache_poisoning.rs#L252 would notice use of `release-please-action` in this workflow. Perhaps because the `setup-node` is in a separate job in the same workflow file. /shrug)

package.json
```diff
@@ -66,16 +66,19 @@
...
     "packageManager": {
       "name": "npm",
-      "version": "^11.9.0",
+      "version": "^11.16.0",
       "onFail": "error"
     },
     "runtime": {
       "name": "node",
-      "version": ">=24"
+      "version": ">=26.3.0"
```
trentm:
```
% node --version
v24.16.0

% npm --version
11.16.0

% npm run
npm error code EBADDEVENGINES
npm error EBADDEVENGINES The developer of this package has specified the following through devEngines
npm error EBADDEVENGINES Invalid devEngines.runtime
npm error EBADDEVENGINES Invalid semver version ">=26.3.0" does not match "v24.16.0" for "runtime"
npm error EBADDEVENGINES {
npm error EBADDEVENGINES   current: { name: 'node', version: 'v24.16.0' },
npm error EBADDEVENGINES   required: { name: 'node', version: '>=26.3.0' }
npm error EBADDEVENGINES }
```

Harsh. :)  
I guess this is fine for browser dev, because no need to support dev/test/whatever on older Node.js versions?

.github/workflows/release-please.yml
overbalance: better safe than sorry. I'll add.

package.json
overbalance: Yep. `devEngines` is for contributors. `engines` is for consumers. CI uses `devEngines` which I guess makes sense but is annoying in that it prevents running node matrices.

overbalance: Also, the reason I bumped both is because while older nodes work fine for development in this repo, 26.3.0 is the first release with npm 11.16.0. I thought it would be less confusing.

EDIT: "runtime" refers to the floor for tooling and since 24 is LTS, I changed my mind and am including it. if you install 24 and use the npm version it includes, installing will fail but it's more honest about requirements.

trentm (approved):

PR #317, opened 2026-06-12
refactor(sdk): move sources into src/initializer folder

overbalance:

The `@opentelemetry/browser-sdk` package keeps all of its sources flat in `src/`, which leaves no room to organize the upcoming exporter and processor components. This PR moves the existing SDK initialization sources into `src/initializer/` so that sibling folders can grow around them, mirroring the per-folder layout already used by the instrumentation package. The name `initializer` follows the term proposed at the Browser SIG for the folder holding the `start*` entry points.

- Moved all initialization files from `packages/sdk/src/` into `packages/sdk/src/initializer/` (pure rename, no content changes; relative imports are unchanged since the files moved together)
- Updated the tsdown entry list to build only the three public entry points (`logs.ts`, `traces.ts`, `start.ts`) instead of globbing every source file, so `dist/` now contains exactly what the `exports` map exposes plus shared chunks

## Type of change

- [x] Internal refactor (non-breaking, no public API change; the `exports` map and dist entry filenames are unchanged)

## How Has This Been Tested?

- [x] `npm run check-types` passes
- [x] `npm run build` passes, including the attw and publint checks; `dist/logs.js`, `dist/traces.js`, and `dist/start.js` are produced at the same paths as before
- [x] `npm test` passes (57 tests in 9 files)

## Checklist:

- [ ] Unit tests have been added (not applicable, file moves and build config only)
- [ ] Documentation has been updated (not applicable)


david-luna:
> This PR moves the existing SDK sources into src/sdk/ so that src/exporters/ and src/processors/ can be added as sibling folders, mirroring the per-folder layout already used by the instrumentation package.

In my mental model processors and exporters are SDK components. Maybe because I see the SDK specs describing them like https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-processor

Do we want this package to hold something which is not part of the SDK? Then we are still in time to discuss again about the name of the package.

Or maybe are the entry points (logs.ts, traces.ts and start.ts) to broad in scope that require a `sdk` category? Maybe the following layout makes sense? (don't mind the names on the second level, just for layout purposes)

```
src
├─┬ processosrs
│ ├── batch
│ └── simple
├─┬ exporters
│ ├── otlp-http
│ └── console
├─┬ propagators
│ ├── trace-context
│ └── baggage
├─┬ samplers
│ ├── probability
│ └── always-on
└─┬ start (or entrypoint?)
  ├── logs      <- exports startLogsSdk
  ├── traces    <- exports startTracesSdk
  └── browser   <- exports startBrowserSdk
```

overbalance:
@david-luna What do you think about `sdks`? or `initializers`? I realize `initializer` isn't a term used in the js repo. @martinkuba would be interested in your feedback too

martinkuba:
The sdk.ts, start.ts, logs.ts, and traces.ts files all seem to have just initialization/configuration. We could move these to a config directory (or similar). We could also leave one in the root directory (like start.ts or initialize.ts) to make it obvious as the entry point. But the other ones basically exist in support of this.

david-luna:
@overbalance 
yesterday we commented about this PR in the Browser SIG. The term `initializer` was shared in the options @martinkuba and I proposed. I think we could go with that name. Also knowing that we will define the export paths in this package there is no problem in review/rename the folder again when we see more clearly which category/name we can give to the `start*` functions.

martinkuba (approved):

Issue #313, opened 2026-06-11
Move web-common package from JS core to this repo


martinkuba:
Sub-issue of https://github.com/open-telemetry/opentelemetry-browser/issues/209

Move components from the `@opentelemetry/web-common` package from [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/web-common).

## Checklist

- [x] Move code to this repo
- [x] Update the `browser-sdk` README with information from web-common README
- [ ] Deprecate the js package
	- [ ] Update the js package README with deprecation notice and migration instructions, publish a final version
	- [ ] Run `npm deprecate` on `@opentelemetry/web-common` pointing to the new location
	- [ ] Delete the `web-common` package source from opentelemetry-js

mquentin:
Hey, first two can be marked done (unless I missed something)

Issue #209, opened 2026-03-20
Migrate or deprecate browser packages from opentelemetry-js and opentelemetry-js-contrib


martinkuba:
Browser-specific packages are currently spread across three repositories: [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js), [opentelemetry-js-contrib](https://github.com/open-telemetry/opentelemetry-js-contrib), and this repository ([opentelemetry-browser](https://github.com/open-telemetry/opentelemetry-browser)). This makes it harder to maintain a coherent browser observability story and creates confusion for users about where to find browser-specific packages.

The goal of this effort is to consolidate all browser-specific packages into this repository, deprecating the original packages and providing clear migration paths. This includes both instrumentations and supporting infrastructure (SDK components, context managers, resource detectors, and shared utilities).

### Event-based instrumentations (straightforward move)

These are newer, event-based instrumentations that can be added directly to the `@opentelemetry/browser-instrumentation` package as subpath exports.

- [`instrumentation-browser-navigation`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-browser-navigation) from contrib
- [`instrumentation-web-exception`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-web-exception) from contrib

### Span-based instrumentations (important, needs migration planning)

These are widely used, span-based instrumentations that require careful migration planning to minimize disruption for existing users.

- [`instrumentation-fetch`](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-fetch) from opentelemetry-js
- [`instrumentation-xml-http-request`](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-xml-http-request) from opentelemetry-js

### Span-based instrumentations (needs discussion)

These are more experimental, span-based instrumentations. Each one needs individual discussion about whether to move, deprecate, or replace with a new implementation.

- [`instrumentation-document-load`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-document-load) — overlaps with navigation-timing
- [`instrumentation-long-task`](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/packages/instrumentation-long-task) — Long Task API being superseded by Long Animation Frames API
[...]

PR #321, opened 2026-06-18
refactor(sdk): migrate @opentelemetry/web-common session processors to browser-sdk/session

mquentin:

First step in migrating [`@opentelemetry/web-common`](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/web-common) from `opentelemetry-js` into this repo. Ports `SessionLogRecordProcessor` and `SessionSpanProcessor` into a @opentelemetry/browser-sdk/session

## Fixes (209 and 313)

https://github.com/open-telemetry/opentelemetry-browser/issues/313
https://github.com/open-telemetry/opentelemetry-browser/issues/209

- New `packages/sdk/src/session/` module: `SessionLogRecordProcessor`, `SessionSpanProcessor`, `SessionProvider` type, local `ATTR_SESSION_ID` constant.
- Exposed via new subpath export `@opentelemetry/browser-sdk/session`.
- add the session processors in the local dev sandbox

Source is functionally identical to upstream; only imports were adapted to repo conventions

## Type of change

## QA

run the sandbox locally `npm run dev` and check the the sessions logs and span processors works as intended as have a `session.id` attribute:

Screenshot 2026-06-18 at 11 41 20

Screenshot 2026-06-18 at 11 57 48

## Notes

This PR https://github.com/open-telemetry/opentelemetry-browser/pull/317 will probably impact this one once approved and merges 👀 on its status


sandbox/src/utils/session-provider.ts
```diff
@@ -0,0 +1,20 @@
+// session-provider.ts — Minimal mutable SessionProvider used by the sandbox.
+// Rotating the id lets QA observe `session.id` changing on subsequently
+// emitted spans and log records.
+
+import type { SessionProvider } from '@opentelemetry/browser-sdk/session';
+
+export interface MutableSessionProvider extends SessionProvider {
```
mquentin: just a way to QA that mutating the current session.id works properly

packages/sdk/src/session/types/SessionProvider.ts
david-luna, line 1: I think this could go on a `types.ts` file. We can split afterwards if the file grows too big

packages/sdk/tsdown.config.ts
```diff
@@ -3,5 +3,5 @@ import baseConfig from '../../tsdown.config.ts';
 
 export default defineConfig({
   ...baseConfig,
-  entry: ['src/*.ts', '!src/*.test.ts'],
+  entry: ['src/*.ts', '!src/*.test.ts', 'src/session/index.ts'],
```
david-luna: Since we expect to have sibling folders to `session`. Could we use a glob here like `src/*/*.ts` for future additions? And of course the glob to exclude tests in subfolders :)

david-luna (commented):
Looks good!!! I wonder if the SDK can provide a `SessionProvider` implementation. this can be done in a follow up PR

mquentin:
> SessionProvider

@david-luna So it was my bad, I missed the migration of the `SessionManager.ts` from [@opentelemetry/web-common](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/web-common)which is the equivalent of my tmp sandbox `SessionProvider`.

I therefore added another commit that migrate the remaining files and remove my added `SessionProvider`.

packages/sdk/src/session/types/SessionProvider.ts
mquentin: Outdated since I also migrated the `SessionManager` from [@opentelemetry/web-common](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/web-common)

packages/sdk/tsdown.config.ts
mquentin: good point ! I've pushed your suggestion

packages/sdk/src/session/SessionManager.ts
```diff
@@ -0,0 +1,176 @@
...
+    const session: Session = {
+      id: sessionId,
+      startTimestamp: Date.now(),
+    };
+
+    // this is async call, but we don't wait for it
+    void this._store.save(session);
+
+    for (const observer of this._observers) {
+      observer.onSessionStarted(session, this._session ?? undefined);
```
david-luna: I think this is something we did not catch when this code was in the web-common package.

Is this actually reporting the new started session? I see the assignment to the class property after this loop. Also the test is only checking for the number of sessions but not if the session ids are the right ones.

packages/sdk/src/session/SessionManager.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    expect(sessionId).toBe('session-2');
+  });
+
+  it('should notify observers when session starts and ends', () => {
+    sessionManager = new SessionManager(config);
+    sessionManager.addObserver(observer);
+
+    sessionManager.getSessionId(); // Starts session-1
+    sessionManager.getSessionId(); // Reuse session-1
+    expect(observer.startedSessions.length).toBe(1);
```
david-luna, line 169: I think here we should also check the right session is being notified as started

```suggestion
    expect(observer.startedSessions.length).toBe(1);
    expect(observer.startedSessions[0].id).toEqual('session-1');
```

packages/sdk/src/session/SessionManager.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    expect(observer.endedSessions.length).toBe(0);
+
+    (sessionManager as unknown as PrivateSessionManager).resetSession();
+    expect(observer.startedSessions.length).toBe(2);
+    expect(observer.endedSessions.length).toBe(1);
```
david-luna, line 178: Same here with started and ended sessions.

david-luna (changes requested):
Sorry @mquentin I need you to fix something that we didn't catch earlier.

packages/sdk/src/session/SessionManager.ts
mquentin: So I did the QA with a unique session with the sandbox so far so I've missed this bug (I also did not check each of the logs (to see if the first one would miss the Id).

I will move `this._session = session;` before and have new `observer.onSessionStarted(session,previousSession);`

david-luna (approved):
thank you!

martinkuba (approved):

martinkuba:
The examples from the README should be moved as well, but this can be handled separately.

mquentin:
> The examples from the README should be moved as well, but this can be handled separately.

@martinkuba I will work on this in a following PR but good point

PR #326, opened 2026-06-24
feat(sandbox): add exception event button

hugolevyy:
## Summary

- Registers `ErrorsInstrumentation` in the sandbox SDK init so unhandled errors are automatically captured as exception log events
- Adds a new **Events** section to the actions panel with a single `💥 Exception` button that throws an unhandled error via `setTimeout`, triggering the instrumentation and producing a log record with `exception.type`, `exception.message`, and `exception.stacktrace`

## How it works

The button throws from inside a `setTimeout` callback so the error escapes React's event handling and reaches `window.onerror`, where `ErrorsInstrumentation` captures it and emits an OTel log record with `eventName: "exception"`.

## In the UI
image

## Payload sent
image

### DIFF
--- a/sandbox/src/app/actions.ts
+++ b/sandbox/src/app/actions.ts
@@ -67,7 +67,9 @@ export function createActions(tracer: Tracer, logger: Logger) {
 
   const exceptionEvent = () => {
     setTimeout(() => {
-      throw new Error('Demo exception event — captured by ErrorsInstrumentation');
+      throw new Error(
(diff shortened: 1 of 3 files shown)


david-luna (approved):
Thank you very much!!!

david-luna:
leaving it unmerged for now to give time to other @open-telemetry/browser-approvers in other timezones to have a look

joaquin-diaz (approved):

PR #329, opened 2026-06-25
feat: add NetworkContextManager proposal for cross-instrumentation OTel context sharing

joaquin-diaz:
## Summary

This is a **proposal PR** to discuss an approach for sharing OTel context between the network (fetch/XHR) and resource timing instrumentations. See the linked issue for the full design discussion.

### Problem

The resource timing instrumentation observes all `PerformanceResourceTiming` entries and emits `browser.resource_timing` logs, but those logs carry no trace context. This makes it impossible to correlate them with the fetch/XHR span that triggered the request.

### Approach

- Add a `NetworkContextManager` utility class (`src/utils/NetworkContextManager.ts`) that stores span contexts keyed by URL + timing window.
- The network instrumentation (fetch/XHR) calls `register(span, url, startPerfNow, endPerfNow)` when a span ends — keeping it decoupled via a generic `onNetworkSpanEnd` callback (not implemented in this PR, as fetch is still on a separate branch).
- The resource timing instrumentation calls `getNetworkContextManager().getContext(entry)` when emitting each log, attaching the matching span context if one exists.
- A `getNetworkContextManager()` singleton makes wiring zero-config for the common case.

### Changes

- `src/utils/NetworkContextManager.ts` — new `NetworkContextManager` class + `getNetworkContextManager()` singleton
- `src/utils/index.ts` — exports the new utility
- `src/resource-timing/instrumentation.ts` — `_emitResource` now passes `context: getNetworkContextManager().getContext(entry)` to `logger.emit`

### What's not in this PR

The fetch instrumentation side (`onNetworkSpanEnd` hook + calling `register()`) lives on the [fetch branch](https://github.com/embrace-io/opentelemetry-browser/tree/joaquin-diaz/feat/migrate-instrumentation-fetch) and will be wired up once that branch merges.

## Test plan

- [ ] Review the `NetworkContextManager` design and timing-window matching logic
- [ ] Confirm the singleton approach is acceptable vs. explicit instantiation
- [ ] Discuss whether the resource timing instrumentation should call `getNetworkContextManager()` directly or via an injected dependency


packages/instrumentation/src/utils/NetworkContextManager.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Stores OTel span contexts keyed by URL and timing window, so that
+ * resource timing entries produced by network requests can be correlated
+ * with the corresponding span.
+ *
+ * The network instrumentation (fetch, XHR) calls `register()` when a span
+ * ends; the resource timing instrumentation calls `getContext()` when
+ * emitting a log for a resource timing entry.
+ */
+export class NetworkContextManager {
```
mquentin: Quick suggestion: We will probably want to have other contexts as other instrumentations will also share contexts.

Could we have `ContextManager<T>` generic and have a framework for later shared context ?

packages/instrumentation/src/utils/index.ts
```diff
@@ -5,4 +5,8 @@
 
 export { getElementCSSSelector } from './getElementCSSSelector.ts';
 export { getElementXPath } from './getElementXPath.ts';
+export {
+  getNetworkContextManager,
+  NetworkContextManager,
```
david-luna: I think exporting the function is enough for now. We only need a singleton

packages/instrumentation/src/utils/NetworkContextManager.ts
david-luna: naming is hard 😓 

`ContextManager` is a concept already used in the API package ([ref](https://github.com/open-telemetry/opentelemetry-js/blob/53337962f2506e2422196b532cb058a533f0b5e3/api/src/context/types.ts#L38)) so I think its better to not reuse it here.

`ContextManager` manages the context across different function boundaries and this class does something different. It stores or registers (the method name is revealing) so I'd dare to propose `NetworkContextStore` or `NetworkContextRegistry`.

We could make it generic if the method signature accepts an options parameter instead of positional ones.

```js
interface ContextRegistry<RegisterOptions, QueryOptions> {
  register(options: RegisterOptions): undefined;
  getContext(query: QueryOptions): Context | undefined
}
```

then the `NetworkContextRegistry` would implement such interface passing its types. Then the getter method should work as a factory (of singletons?). Is not straight forward

Given that we will keep this method and class internal for now I would prefer to discuss this when we have a real situation.

packages/instrumentation/src/utils/NetworkContextManager.ts
```diff
@@ -0,0 +1,70 @@
...
+   * respectively — the same reference frame as `PerformanceResourceTiming`
+   * timestamps.
+   */
+  register(
+    span: Span,
+    url: string,
+    startPerfNow: number,
+    endPerfNow: number,
+  ): void {
+    const ctx = trace.setSpan(ROOT_CONTEXT, span);
```
david-luna: should we be defensive here and check if the consumer tries to register the same span twice?

packages/instrumentation/src/utils/index.ts
joaquin-diaz: Is exposed for testing, as long as we don't expose it to the consumer in the instrumentation exports should be fine

packages/instrumentation/src/utils/NetworkContextManager.ts
joaquin-diaz: We might have the same URL being registered multiple times if there are multiple calls to the same endpoint. That's why `NetworkContextRegistry` also uses timestamps to get the right context. This is the register function for the base class:

```
  register(span: Span, data: TData): void {
    const ctx = trace.setSpan(ROOT_CONTEXT, span);
    const list = this._records.get(data.key) ?? [];
    list.push({ ...data, ctx });
    this._records.set(data.key, list);
  }
  ```
  
It allows you to push multiple contexts to the same key. For network requests that's the URL, for any other registry it might be something different. If repetition is not actually possible then the class implementing ContextRegistry only needs to return the first entry by key

**Comment by joaquin-diaz** [comment]
@mquentin @david-luna thanks for the review! I updated the PR

File: packages/instrumentation/src/utils/ContextRegistry.ts [hunk 4]
```diff
@@ -0,0 +1,22 @@
...
+type StoredRecord<TData> = TData & { ctx: Context };
+
+export abstract class ContextRegistry<TData extends { key: string }, TLookup> {
+  protected _records = new Map<string, StoredRecord<TData>[]>();
+
+  register(span: Span, data: TData): void {
+    const ctx = trace.setSpan(ROOT_CONTEXT, span);
+    const list = this._records.get(data.key) ?? [];
+    list.push({ ...data, ctx });
+    this._records.set(data.key, list);
```
david-luna, line 33: The record is set but never removed. I wonder if this can lead to allocating too many memory in some scenarios. For example a SPA that does many requests and it's working for a long period of time.

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { Context, Span } from '@opentelemetry/api';
+import { ROOT_CONTEXT, trace } from '@opentelemetry/api';
+
+type StoredRecord<TData> = TData & { ctx: Context };
+
+export abstract class ContextRegistry<TData extends { key: string }, TLookup> {
+  protected _records = new Map<string, StoredRecord<TData>[]>();
+
+  register(span: Span, data: TData): void {
+    const ctx = trace.setSpan(ROOT_CONTEXT, span);
```
david-luna, line 30: [question] is it okay to register the same span twice? I guess in some situations it is but not completely sure for the case of `NetworkContextRegistry`

david-luna (commented):
@joaquin-diaz this looks very good. Just a couple of doubts and I think we're good to go :)

packages/instrumentation/src/utils/ContextRegistry.ts
joaquin-diaz: Do you think we should timeout entries after a fixed amount of time?

packages/instrumentation/src/utils/ContextRegistry.ts
joaquin-diaz: Today I'd say no but I don't know how strict we want to be in the base class without knowing where else we could use this. Should we wait and see if many consumers of the base class end up validating for duplicated spans?

david-luna: > Should we wait and see if many consumers of the base class end up validating for duplicated spans?

yes :)

packages/instrumentation/src/utils/ContextRegistry.ts
david-luna: I think I'm biased because fetch/xhr instrumentations only need the the context (to match a resource timing entry) for a while. But I cannot think on a use case where wee need the context to say permanently 🤔 

I think you made a valid point on not being very strict on the base class. So if we implement a cleanup process it should be done in the specific class

Here is a super simple implementation
```js
const TIME_TO_LIVE = 1_000; // fetch/xhr wait around 300ms to query the resource timings
export class NetworkContextRegistry extends ContextRegistry<TData, TLookup> {
  register(span: Span, data: TData) {
    super.register(span, data);
    setTimeout(() => this._records.delete(data.key), TIME_TO_LIVE);
  }
}
```
This is just for example purposes. I think we could avoid to create a timer and a closure for every registration with another pattern (clear map when idle?).

joaquin-diaz: I took a different approach and added an `unregister` method so we don't need to make any assumptions on how long to hold context and we shift that responsibility up. For example:

```
// resource-timing.ts
private _emitResource(entry: PerformanceResourceTiming): void {
...
const ctx = networkContextRegistry.getContext(entry)
this._logger.emit({ctx, ...})
networkContextRegistry.unregister(entry)
```

What do you think?

packages/instrumentation/src/utils/NetworkContextRegistry.ts
```diff
@@ -0,0 +1,36 @@
...
+import type { Context } from '@opentelemetry/api';
+import { ContextRegistry } from './ContextRegistry.ts';
+
+export interface NetworkSpanData {
+  key: string;
+  startPerfNow: number;
+  endPerfNow: number;
+}
+
+export class NetworkContextRegistry extends ContextRegistry<
```
mquentin, line 15: >  extends ContextRegistry<
  NetworkSpanData,
  PerformanceResourceTiming
>

Thanks for taking my request in consideration.  That's exactly what I was suggesting, this approach will help further context definition.

mquentin (approved):
LGTM, I find the naming pretty self-explanatory and the generic class tests also help document the overall implementation.

david-luna (approved):

PR #332, opened 2026-07-02
feat: add e2e test infrastructure with ErrorsInstrumentation example

joaquin-diaz:
## Summary

- Adds an `e2e-tests/` workspace with a Vitest + Playwright (Chromium) browser test runner
- Introduces shared test utilities: `test-collector.ts` (intercepts OTLP exports via MSW) and `test-otel-setup.ts` (wires up trace/log providers and registers instrumentations per test)
- Adds `ErrorsInstrumentation` e2e tests as a working example covering Error objects, string errors, unhandled rejections, and `applyCustomAttributes`
- Adds `test:e2e` script to the root `package.json` and hooks it into CI

### DIFF
--- a/e2e-tests/packages/instrumentation/errors/instrumentation.test.ts
+++ b/e2e-tests/packages/instrumentation/errors/instrumentation.test.ts
@@ -1,0 +1,176 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ErrorsInstrumentation } from '@opentelemetry/browser-instrumentation/experimental/errors';
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  describe,
+  expect,
+  it,
+  vi,
+} from 'vitest';
+import { startMsw, stopMsw } from '../../../utils/test-collector.ts';
+import type { TestOtelSetupResult } from '../../../utils/test-otel-setup.ts';
+import { testOtelSetup } from '../../../utils/test-otel-setup.ts';
+
+// Dispatch a synthetic error event without actually crashing the browser page.
+// We suppress the default reporting behavior with a capture-phase listener that
+// calls preventDefault() so Playwright does not treat it as a test failure.
+const dispatchErrorEvent = (error: Error | string) => {
+  const event = new Event('error', { cancelable: true });
+  Object.defineProperty(event, 'error', { value: error });
+  const suppress = (e: Event) => e.preventDefault();
+  window.addEventListener('error', suppress, { capture: true });
+  try {
+    window.dispatchEvent(event);
+  } finally {
+    window.removeEventListener('error', suppress, { capture: true });
+  }
+};
+
+const dispatchUnhandledRejection = (reason: Error | string) => {
+  const event = new PromiseRejectionEvent('unhandledrejection', {
+    promise: Promise.resolve(),
+    reason,
+    cancelable: true,
+  });
+  window.dispatchEvent(event);
+};
+
+describe('ErrorsInstrumentation', () => {
+  let result: TestOtelSetupResult;
+
+  beforeAll(async () => {
+    await startMsw();
+  });
+
+  afterAll(() => {
+    stopMsw();
+  });
+
+  afterEach(async () => {
+    await result.cleanup();
+  });
+
+  it('emits a log record with exception attributes when an Error is thrown', async () => {
+    result = testOtelSetup([new ErrorsInstrumentation()]);
+
+    const error = new TypeError('Something went wrong');
+    dispatchErrorEvent(error);
+
+    await vi.waitFor(
+      () => {
+        const logs = result.getLogs();
+        expect(logs).toHaveLength(1);
+        const log = logs[0];
+        expect(log).toBeDefined();
+        if (!log) {
+          return;
+        }
+
+        const attr = (key: string) =>
+          log.attributes.find((a) => a.key === key)?.value;
+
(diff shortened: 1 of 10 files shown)


e2e-tests/utils/test-otel-setup.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export function testOtelSetup(
+  instrumentations: Instrumentation[],
+): TestOtelSetupResult {
+  const collector = setupCollector();
+
+  const traceExporter = new OTLPTraceExporter({
+    url: COLLECTOR_URL,
+    timeoutMillis: 1,
```
overbalance: msw instantly processes the body so I think you can safely remove this
```suggestion
```

e2e-tests/utils/test-otel-setup.ts
```diff
@@ -0,0 +1,65 @@
...
+  });
+  const provider = new WebTracerProvider({
+    spanProcessors: [new SimpleSpanProcessor(traceExporter)],
+  });
+  provider.register();
+
+  const logExporter = new OTLPLogExporter({
+    url: LOGS_COLLECTOR_URL,
+    timeoutMillis: 1,
```
overbalance:
```suggestion
```

e2e-tests/utils/test-collector.ts
```diff
@@ -0,0 +1,123 @@
...
+        ),
+      ),
+    getLogs: () =>
+      logPayloads.flatMap((p) =>
+        p.resourceLogs.flatMap((rl) =>
+          rl.scopeLogs.flatMap((sl) => sl.logRecords),
+        ),
+      ),
+    cleanup: () => {
+      worker?.resetHandlers();
```
david-luna: should we empty here the payload arrays?

e2e-tests/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
...
+        fileURLToPath(
+          new URL(
+            '../packages/instrumentation/src/errors/index.ts',
+            import.meta.url,
+          ),
+        ),
+    },
+  },
+  test: {
+    onConsoleLog: () => (process.env['VERBOSE'] ? undefined : false),
```
overbalance: This suppresses all console output unless VERBOSE is set, including MSW `onUnhandledRequest: 'error'` and OTel diag errors. If an export fails we would see only a timeout with no reason. If we know the level, can we let error logs through always, and keep the VERBOSE gate for the rest?

e2e-tests/packages/instrumentation/errors/instrumentation.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    const error = new TypeError('Something went wrong');
+    dispatchErrorEvent(error);
+
+    await vi.waitFor(
+      () => {
+        const logs = result.getLogs();
+        expect(logs).toHaveLength(1);
+        const log = logs[0];
+        expect(log).toBeDefined();
+        if (!log) {
```
overbalance, line 71: Since `toHaveLength(1)` already throws when the array is empty, this `if (!log) return;` never runs today. But if someone later relaxes the length check, the early return makes `vi.waitFor` resolve as success without running the attribute assertions. Could we throw here instead of returning, so the test cannot silently pass?

e2e-tests/packages/instrumentation/errors/instrumentation.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    cancelable: true,
+  });
+  window.dispatchEvent(event);
+};
+
+describe('ErrorsInstrumentation', () => {
+  let result: TestOtelSetupResult;
+
+  beforeAll(async () => {
+    await startMsw();
```
david-luna: I think we could use other names to avoid other implementers to think about infra details of the test.

I think i the concept of a collector is a good idea. If we expand its interface to have a stop method we could hide MSW details in it. Also it will separate concerns. Collector file contains everything related to start/stop and inspecting received data. Otel setup file just handles the registration of providers and instrumentations.

```js
import { setupCollector } from '../../../utils/test-collector.ts';
import { setupOtelSdk } from '../../../utils/test-otel-setup.ts';
const collector = setupCollector();

beforeAll(()=> collector.start());
afterAll(()=> collector.stop());
afterEach(()=> collector.cleanup());

// ... in the test
describe(() => {
  let sdk;
  
  afterEach(async () => {
    if (sdk) {
      await sdk.shutdown();
      sdk = undefined;
    }
  });
  it('...', () => {
    sdk = setupOtelSdk([new ErrorsInstrumentation()]);
    // test body
  });
});
```

e2e-tests/utils/test-otel-setup.ts
joaquin-diaz: This is the time the batch exporter waits to export data, otherwise the test would have to "wait" 10s, and we would have to setup a fake clock in each test to actually avoid waiting that time

e2e-tests/packages/instrumentation/errors/instrumentation.test.ts
joaquin-diaz: Yeah good call

e2e-tests/vitest.config.ts
joaquin-diaz: I don't see a way of filtering unless assuming all errors would contain "error" or a similar string

```
onConsoleLog?: (log: string, type: "stdout" | "stderr", entity: TestModule | TestCase | TestSuite | undefined) => boolean | void;
```

I'll remove the check for now and we can add it back if it gets too noisy

overbalance (approved):

PR #333, opened 2026-07-02
refactor(sdk): restructure exports into per-export folders with barrels

overbalance:

After #317 the SDK had an inconsistent layout. Public entrypoints (logs, traces, start) sat as bare files at the dist root while their internals lived under `initializer/`, and `session/` was the only export with its own folder and barrel. The `exports` map mixed conventions and did not surface the public config types.

- Give every export its own folder with a thin `index.ts` barrel, so all subpaths resolve uniformly to `dist/<name>/index.js` (`.`, `./logs`, `./traces`; `./session` already had this shape).
- Move shared internals (`types.ts`, `diag.ts`, `sdk.ts`) and the unused modules (`context.ts`, `sampler.ts`, `propagation.ts`) into `src/core/`. Implementation files are renamed to match their export and moved with `git mv` to preserve history.
- Re-export the public config types from each barrel (`QuickStartConfig`, `RootConfig`, `LogsConfig`, `TracesConfig`, `CommonConfig`, `ExportConfig`, `BatchProcessorConfig`, `WebSdk`). These were not exported before.
- Stop exporting `WebSdkFactory`. It is only used by `combineSdks`, so it becomes a local type in `core/sdk.ts`.
- Rewrite the `exports` map to explicit `types`/`default` conditions, add a `./package.json` entry, and update the tsdown `entry` map.

Public subpaths are unchanged, so this is not breaking for consumers.

## Type of change

## How Has This Been Tested?

- [x] `npm run build` for `@opentelemetry/browser-sdk`: attw and publint pass
- [x] `npm run test` for `@opentelemetry/browser-sdk`: existing tests pass (moved with their sources, no behavior change)
- [x] tsc, eslint and biome via the pre-commit hook

## Checklist:

- [ ] Unit tests have been added (no new tests; existing tests moved with their sources, no behavior change)


david-luna (approved):
@overbalance thanks for this. I guess when we add more components here new folders will flourish. And maybe some of the parts in core will have their own space.

this is a good start :)

overbalance:
I envision "./core" as temporary space for non-exported files

martinkuba (approved):

martinkuba:
I like this new structure 👍

PR #335, opened 2026-07-03
fix(sandbox): fix jsError action to actually throw

hugolevyy:
## Summary

The `💥 JS Error` button in the sandbox Traces section was a silent no-op.

`void {}.undefinedProperty` does not throw in JavaScript — accessing a missing property on a plain object returns `undefined`. As a result, the `catch` block never ran, `span.recordException()` was never called, and the span was always exported with a healthy status despite the button's intent.

Replaced with `throw new Error(...)` so the catch block actually executes, `recordException` is called with a real `Error` object, and the span is correctly exported with `exception.type`, `exception.message`, `exception.stacktrace` and status `ERROR`.

The `@ts-expect-error` directive is also removed — it was suppressing a TypeScript type error, which is unrelated to a JavaScript runtime exception.


sandbox/src/app/actions.ts
```diff
@@ -50,8 +50,7 @@ export function createActions(tracer: Tracer, logger: Logger) {
   const jsError = () => {
     const errorSpan = tracer.startSpan('js-error-event');
     try {
-      // @ts-expect-error generating an error on purpose
```
overbalance, line 53: Would you mind refactoring it to different invalid code without an explicit Error?

overbalance (approved):

PR #339, opened 2026-07-07
feat:proposal(sdk): use crypto.randomUUID() for session ID generation

hugolevyy:
Proposal to replace the custom `Math.random()`-based hex generator in `DefaultIdGenerator` with `crypto.randomUUID()`, producing standard UUID v4 session IDs.

This also switches to a cryptographically secure source of randomness and removes ~15 lines of custom ID generation code.


overbalance:
Thanks for looking into this. This was previously tried and it's sloooooooow: https://github.com/open-telemetry/opentelemetry-js/pull/6209#discussion_r2611342269

PR #340, opened 2026-07-09
fix(sdk): pass config to startBrowserSdk in quickStartBrowserSdk

martinkuba:
`quickStartBrowserSdk` built an sdkConfig object from the user's options but called `startBrowserSdk()` without passing it, so every option was silently dropped and the SDK always started with defaults.

This is a one-line fix + tests.

### DIFF
--- a/packages/sdk/src/startBrowserSdk.test.ts
+++ b/packages/sdk/src/startBrowserSdk.test.ts
@@ -5,12 +5,43 @@
 
 import { diag, trace } from '@opentelemetry/api';
 import { logs } from '@opentelemetry/api-logs';
-import { afterAll, afterEach, describe, expect, it, vi } from 'vitest';
+import type { MockInstance } from 'vitest';
+import {
+  afterAll,
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  vi,
+} from 'vitest';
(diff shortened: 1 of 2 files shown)


overbalance (approved):

packages/sdk/src/startBrowserSdk.test.ts
```diff
@@ -127,3 +158,105 @@ describe('startBrowserSdk', () => {
...
+    ).toBeDefined();
+    fetchSpy.mock.calls.forEach((args) => {
+      expect(args[1]).containSubset({
+        method: 'POST',
+        headers: { bar: 'baz' },
+      });
+    });
+  });
+
+  it('should propagate service name and version as resource attributes', async () => {
```
mquentin, line 249: I see that `quickStartBrowserSdk` also provide a `disabled` config parameter. Could you add a test to also cover this use case ? It would be great to prevent regression if users implement an opt out mechanism.

mquentin (commented):
Small test suggestion

packages/sdk/src/startBrowserSdk.test.ts
martinkuba: I have added the test

mquentin (approved):
LGTM

PR #341, opened 2026-07-09
chore(sdk): enable publishing for browser-sdk

martinkuba:
Remove the `private` flag so the package can be published, and wire packages/sdk into release-please (config, manifest, and the CI publish step).

The first publish will be manual, so that trusted publisher can be configured. After that, release please should pick up subsequent releases.


mquentin (approved):
LGTM ! great news

overbalance (approved):

PR #346, opened 2026-07-10
refactor(sandbox): migrate SDK setup to official startBrowserSdk

mquentin:

Replace the hand-wired `WebTracerProvider/LoggerProvider` setup in the
sandbox with the official `@opentelemetry/browser-sdk` startBrowserSdk
entry point, keeping the instrumented behavior identical:

- session processors run first, then console + optional UI-mirror
  exporters, via traces.processors / logs.processors
- batching OTLP export preserved through exportConfig +
  batchProcessorConfig (maxExportBatchSize 10, scheduledDelayMillis 1000)
- StackContextManager + W3C trace-context/baggage propagators passed
  explicitly to reproduce WebTracerProvider.register() defaults, so
  fetch/XHR context propagation and header injection keep working
- registerInstrumentations block unchanged

## QA

`npm run dev` is still showcasing that the SDK is still loaded properly and reporting the starting the same instrumentations.

`github.io` deployment of the sandbox under `https://open-telemetry.github.io/opentelemetry-browser/` will be the followup to QA the change

Screenshot 2026-07-10 at 15 03 05

Decorative code snippet has also been updated but with the goal of not being exhaustive.

Screenshot 2026-07-10 at 15 25 16

## Follow up PR

I plan to work on a small Mock Service Worker like https://www.npmjs.com/package/msw to showcase how payloads are sent to intake. I will however discuss about it with the SIG to see if there is traction about this.


sandbox/package.json
```diff
@@ -12,18 +12,14 @@
     "@opentelemetry/api": "^1.9.1",
     "@opentelemetry/api-logs": "^0.220.0",
     "@opentelemetry/browser-instrumentation": "file:../packages/instrumentation",
-    "@opentelemetry/browser-sdk": "file:../packages/sdk",
+    "@opentelemetry/browser-sdk": "^0.1.0",
```
overbalance: I would revert this. You're going to have a gap between release and when renovate runs, where you're pulling from NPM and not the local copy.

mquentin: Indeed better be on sync with the local version if we did not released the npm package for a while.

sandbox/src/otel.ts
```diff
@@ -58,6 +53,14 @@ interface OtelHandle {
   sessionManager: SessionManager;
 }
 
+// Batch processor settings shared by the traces and logs exporters. Passing
+// this alongside `exportConfig` makes the SDK append a batching OTLP exporter
+// after our custom processors below.
+const BATCH_PROCESSOR_CONFIG = {
```
overbalance: Do we have a `type` for this?

sandbox/src/otel.ts
```diff
@@ -58,6 +53,14 @@ interface OtelHandle {
   sessionManager: SessionManager;
 }
 
+// Batch processor settings shared by the traces and logs exporters. Passing
+// this alongside `exportConfig` makes the SDK append a batching OTLP exporter
+// after our custom processors below.
```
overbalance:
```suggestion
// Batch tuning for the OTLP exporters that startBrowserSdk appends when
// `exportConfig` is set. Shared by the traces and logs signals.
```

sandbox/src/otel.ts
```diff
@@ -66,12 +69,6 @@ export async function initOtel(
   customAttrs: Record<string, string> = {},
   { onSpan, onLog }: InitOtelOptions = {},
 ): Promise<OtelHandle> {
```
overbalance, line 71: we may want to port this into the sdk itself

```suggestion
): Promise<OtelHandle> {

  // ── Validate export endpoints ───────────────────────────────────────────────
  // startBrowserSdk logs a diag.error and silently skips the OTLP exporter for
  // an invalid URL, so guard here and throw to surface the failure through the
  // caller's error handling instead of reporting "SDK ready".
  for (const url of [config.tracesUrl, config.logsUrl]) {
    if (!URL.parse(url)) {
      throw new Error(`Invalid OTLP export URL: "${url}"`);
    }
  }
```

sandbox/src/otel.ts
```diff
@@ -124,11 +105,34 @@ export async function initOtel(
...
-  const logProvider = new LoggerProvider({
-    resource,
-    processors: logProcessors,
+  // ── SDK ─────────────────────────────────────────────────────────────────────
+  // startBrowserSdk registers the tracer and logger providers. For each signal
+  // it keeps the custom processors above and, because `exportConfig` is set,
+  // appends a BatchSpanProcessor / BatchLogRecordProcessor exporting over OTLP.
+  //
+  // The traces `contextManager` and `propagators` reproduce what
+  // `WebTracerProvider.register()` used to wire up by default, so async context
```
overbalance: this one is sync, zone is async

```suggestion
  // `WebTracerProvider.register()` used to wire up by default, so context
```

overbalance (changes requested):

sandbox/src/otel.ts
mquentin: Indeed there is `TracesConfig['batchProcessorConfig']`

sandbox/src/otel.ts
mquentin: So I will add this code in the sandbox first and opened a followup PR, what do you think ?

overbalance: Sounds good!

overbalance (approved):

joaquin-diaz (approved):

PR #353, opened 2026-07-16
test(instrumentation): add console instrumentation e2e test

pjsny:

This adds end-to-end coverage for `ConsoleInstrumentation`, building on the e2e testing initiative from #332. This is test-only, no runtime/source changes.

- Adds `ConsoleInstrumentation` e2e tests (`e2e-tests/packages/instrumentation/console/`)
     driving the real browser `console` through the SDK → OTLP/HTTP → MSW-mocked
     collector pipeline, covering:
     - body, severity number/text, `browser.console` event name, `browser.console.method` attribute
     - severity mapping across all five methods (debug/log/info/warn/error)
     - object-argument JSON serialization
     - `logMethods` config (only configured methods emit)
     - trace↔log correlation (a `console.*` call inside an active span yields a log
       carrying the span's `traceId`/`spanId`)
   - Extends shared `test-collector.ts` OTLP log type with `body`/`eventName` (already on the wire).
   - Adds a `StackContextManager` in `test-otel-setup.ts` so trace context propagates
     (needed for the correlation test); names its return type `TestSdkHandle` and
     migrates the `errors` e2e test onto it.
   - Adds the `console` source alias to `e2e-tests/vitest.config.ts`.

   Assertions match records by body, so any potential future unrelated logs emitted in the same window
   won't cause flakes. 

## Type of change

   - [x] Test-only change (no runtime behavior change)
   
 ## How Has This Been Tested?
   
   `npm run test:e2e`

### DIFF
--- a/e2e-tests/packages/instrumentation/console/instrumentation.test.ts
+++ b/e2e-tests/packages/instrumentation/console/instrumentation.test.ts
@@ -1,0 +1,171 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { context, trace } from '@opentelemetry/api';
+import { ConsoleInstrumentation } from '@opentelemetry/browser-instrumentation/experimental/console';
+import {
+  afterAll,
+  afterEach,
+  beforeAll,
+  describe,
+  expect,
+  it,
+  vi,
+} from 'vitest';
+import { collector } from '../../../utils/test-collector.ts';
+import type { TestSdkHandle } from '../../../utils/test-otel-setup.ts';
+import { testSdkSetup } from '../../../utils/test-otel-setup.ts';
+
+describe('ConsoleInstrumentation', () => {
+  let result: TestSdkHandle;
+
+  beforeAll(async () => {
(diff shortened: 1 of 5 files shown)


e2e-tests/packages/instrumentation/console/instrumentation.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  afterEach(async () => {
+    await result.shutdown();
+    collector.reset();
+  });
+
+  it('emits a log record with body, severity, and method attribute for console.error', async () => {
+    result = testSdkSetup([new ConsoleInstrumentation()]);
+
+    console.error('database connection failed');
```
joaquin-diaz: I know this is just for testing but can you use more browser-like error message just to avoid any kind of confusion on this running in node?

joaquin-diaz (approved):

e2e-tests/packages/instrumentation/console/instrumentation.test.ts
pjsny: Yes. Makes sense. Updated.

PR #357, opened 2026-07-20
feat(sdk): validate all export URLs before starting the SDK

mquentin:
## What

Following this comment of a previous and merged PR:
https://github.com/open-telemetry/opentelemetry-browser/pull/346#discussion_r3587679695

The SDK now validates every OTLP export URL (root **and** per-signal) before
starting — an invalid URL logs a `diag.error` and returns a no-op SDK instead of
silently dropping telemetry.

## Why

Only the root URL was validated. Invalid per-signal URLs just skipped the
exporter, so with custom `processors` the SDK reported success while exporting
nothing. This moves the check into the SDK so the sandbox's manual guard is no
longer needed.

## Changes

- `core/exportUrl.ts` (new): `parseExportUrl()` helper.
- `core/sdk.ts`: validate signal URLs up-front, bail to `NOOP_SDK` before starting
  either signal; unset URLs inherit the root endpoint.
- `traces`/`logs` SDKs: invalid URL returns `NOOP_SDK` instead of silently
  skipping the exporter.
- `sandbox/src/otel.ts`: removed the redundant guard.


packages/sdk/src/core/sdk.ts
```diff
@@ -82,17 +83,32 @@ export function combineSdks<T extends SdkFactories>(
...
+    // signal's provider is registered while the other refuses to start.
+    const endpointUrl = parseExportUrl(
       rootConfig.exportConfig?.url || DEFAULT_OTLP_ENDPOINT,
     );
-
     if (!endpointUrl) {
-      diag.error(
-        `Invalid export URL "${rootConfig.exportConfig.url}". Browser SDK won't start.`,
-      );
       // TODO: need to discuss with the SIG if it's better to return `undefined`
```
mquentin: Propagated the TODO here as the topic still needs to be discussed

packages/sdk/src/core/sdk.ts
```diff
@@ -32,6 +33,17 @@ type ExtractConfigs<T> = Partial<{
...
+/**
+ * Concrete view of the config used inside `startSdk`. It is structurally a
+ * superset of `RootConfig & ExtractConfigs<T>` for any `T`, so the returned
+ * function stays assignable to the precise, factory-derived public type while
+ * the body can read each signal's config without casting.
+ */
+type CombinedConfig = RootConfig & {
+  logs?: RemoveCommonProps<LogsConfig>;
+  traces?: RemoveCommonProps<TracesConfig>;
+};
```
mquentin, line 45: Second commit of the PR ([refactor(sdk): drop signal-config casts in combineSdks](https://github.com/open-telemetry/opentelemetry-browser/pull/357/commits/0d5dc888ea68e9d754779655bb406b6d7d4c4aa3)) goal is to prevent having dangerous casting [here](https://github.com/open-telemetry/opentelemetry-browser/pull/357/changes#diff-5b23a843aafad85ddcf2400838b3792a36c182ba08687850a757183dc4d620f5R110) such as

```
    const signalExportUrls: [string, string | undefined][] = [
      ['Logs SDK', (config?.logs as LogsConfig | undefined)?.exportConfig?.url],
      [
        'Traces SDK',
        (config?.traces as TracesConfig | undefined)?.exportConfig?.url,
      ],
    ];
```

I will bring the point to the SIG

mquentin:
Related to the SIG discussion about enabling Traces if the log URL is the only erroneous URL and enabling the Logs if the trace URL is the only erroneous URL, it makes quite some changes in the SDK:

https://github.com/mquentin/opentelemetry-browser/pull/1

I suggest we do not proceed with such an impactful change and keep it simple for now: if any URL is erroneous we return a fast NOOP_SDK which is the recommended solution of Trent Mick

david-luna:
> Related to the SIG discussion about enabling Traces if the log URL is the only erroneous URL and enabling the Logs if the trace URL is the only erroneous URL, it makes quite some changes in the SDK:
> 
> [mquentin#1](https://github.com/mquentin/opentelemetry-browser/pull/1)
> 
> I suggest we do not proceed with such an impactful change and keep it simple for now: if any URL is erroneous we return a fast NOOP_SDK which is the recommended solution of Trent Mick

For completeness I would add this aligns with the `NodeSDK` behavior with declarative configuration. I also think is a good idea to behave the same way so we do not confuse dev that instrument both frontend and backend

packages/sdk/src/core/sdk.ts
david-luna: I don't fully get here about the dangerous casting. However I'm not opposed to use a stricter type here

The intent of `ExtractConfigs<T>` was to allow users of the `combineSdks` add extra SDKs if they se it fit. So the following code is typed

```js
const startCustomSdk = (config: CustomConfig ) => {
  // return `WebSdk` object
}

// Make SDK composition
const startBrowserSdkWithCustom = combineSdks({
  traces: startTracesSdk,
  logs: startLogsSdk,
  custom: startCustomSdk,
});

// now the combined function has `custom` as property of the config type the user defined
startBrowserSdkWithCustom({
  serviceName: 'foo',
  traces: { /**/ },
  logs: { /**/ },
  custom: { /**/ }, // <- the editor provides type autocompletion for this
});
```

With the `CombinedConfig` we have to maintain it manually and update if necessary.

Now is see the `combineSdks` function is not exported. IMHO this limits the user to only the SDKs we provide here (logs, traces and both). There is always the option of telling users to create a wrapper function around the `start*Sdk` functions.

packages/sdk/src/core/sdk.ts
```diff
@@ -48,7 +60,7 @@ export function combineSdks<T extends SdkFactories>(
 ): WebSdkFactory<RootConfig & ExtractConfigs<T>> {
   // The returned function will transform some of the global
   // configuration options to signal specific ones if the SDK is available
-  return function startSdk(config?: RootConfig & ExtractConfigs<T>) {
+  return function startSdk(config?: CombinedConfig) {
```
david-luna, line 71: If from the conversation above we decide to change the type then line 60 needs to change as well

packages/sdk/src/core/sdk.ts
david-luna: EDIT: the new type makes sense since the combine function only looks for `logs` and `traces` properties. The logic of the method is only resolving the common config properties before calling the factory function. I think this could be done in a loop so any given SDK factory function by the user will be called too.

I guess the question we should ask is. Should we expose the `combineSdks` function?

mquentin: > I don't fully get here about the dangerous casting.

I was talking about each of the `config?.logs as LogsConfig`

> I guess the question we should ask is. Should we expose the combineSdks function?

Could be a good topic for the SIG

packages/sdk/src/core/sdk.ts
mquentin: So I had a look at this, and switching to `CombinedConfig` line 60 will result in a less strict typing for the overall `combineSdks`:

going for

```
export function combineSdks<T extends SdkFactories>(
  factories: T,
): WebSdkFactory<CombinedConfig>
```

would allow such a call no to have a type error:

```
const start = combineSdks({ traces: startTracesSdk });

start({ traces: { exportConfig: { url } } });
start({ logs:   { exportConfig: { url } } });
```

will keeping the current PR version with `WebSdkFactory<RootConfig & ExtractConfigs<T>>`

would prevent you to start the logs as you initial `combineSdks` only has the Trace config.

packages/sdk/src/core/sdk.ts
david-luna: > Could be a good topic for the SIG

Added

packages/sdk/src/core/sdk.ts
david-luna: So `CombinedConifg` is just used internally, now I get it. Thanks!

david-luna (approved):
LGTM. I'd wait a bit for other approvers to give feedback befor merge.

packages/sdk/src/core/exportUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @param scope Human readable name of the SDK used in the error message.
+ * @returns The parsed {@link URL} or `null` when the URL is invalid.
+ */
+export function parseExportUrl(url: string, scope = 'Browser SDK'): URL | null {
+  const parsed = URL.parse(url);
+  if (!parsed) {
+    diag.error(`Invalid OTLP export URL "${url}". ${scope} won't start.`);
+  }
+  return parsed;
+}
```
overbalance, line 32: `URL.parse` only checks that the string can be parsed, not that it points at an HTTP endpoint. `URL.parse('localhost:4318')` succeeds, with protocol `localhost:` and an opaque path, so forgetting `http://` passes this check. It then gets worse further down: because the path is opaque, `endpointUrl.pathname = '/v1/logs'` in `sdk.ts` is silently ignored, so both signals end up posting to `localhost:4318` with no signal path and the SDK still reports a successful start. `javascript:` and `data:` URLs pass too. What do you think about checking the protocol?

```suggestion
export function parseExportUrl(url: string, scope = 'Browser SDK'): URL | null {
  const parsed = URL.parse(url);
  // `URL.parse` accepts any parseable URL, so a missing scheme slips through:
  // `localhost:4318` parses with protocol `localhost:` and an opaque path.
  if (
    !parsed ||
    (parsed.protocol !== 'http:' && parsed.protocol !== 'https:')
  ) {
    diag.error(`Invalid OTLP export URL "${url}". ${scope} won't start.`);
    return null;
  }
  return parsed;
}
```

One related question. `validateUserProvidedUrl` in `otlp-exporter-base` parses with `globalThis.location?.href` as base, so a relative URL like `/v1/traces` is valid for the exporter but rejected here. Do we want to support a same-origin collector path? I think it might be confusing but if we do, this needs a base rather than a plain parse, and that is a slightly different decision from the scheme check.

packages/sdk/src/core/sdk.ts
```diff
@@ -82,21 +94,36 @@ export function combineSdks<T extends SdkFactories>(
     };
 
     const sdks: WebSdk[] = [];
-    const endpointUrl = URL.parse(
+
+    // Validate every export URL up-front and bail out on the first invalid one.
+    // Doing this before starting any signal SDK avoids a partial start where one
+    // signal's provider is registered while the other refuses to start.
```
overbalance: One detail on the comment: a partial start is still possible, it just does not come from a bad URL. If a signal passes `processors: []` the array is truthy, so `exportConfig` is never propagated, no batch processor is added, and `startLogsSdk` bails at its `No LogRecord processors configured` check. By that point traces has already registered its global provider. I ran `startBrowserSdk({ logs: { processors: [] } })` and got the logs error while traces kept exporting normally. Could we scope the comment to the URL case? Moving the empty-processors check up here too would make the original claim true, if you think that is worth doing.

```suggestion
    // Validate every export URL before starting any signal SDK, so an invalid
    // URL cannot leave one signal exporting while the other refuses to start.
```

packages/sdk/src/startBrowserSdk.test.ts
```diff
@@ -103,6 +104,36 @@ describe('startBrowserSdk', () => {
     expect(fetchSpy).not.toHaveBeenCalled();
   });
 
+  it('should not start for an invalid signal-specific URL even with custom processors', async () => {
+    // Arrange: the traces signal provides its own processors and an invalid
+    // export URL — the scenario the sandbox used to guard against by hand.
+    let exportCalled = false;
+
+    // Act
+    browserSdk = startBrowserSdk({
```
overbalance, line 115: I deleted the whole validation loop above (lines 116-123) and this test still passed, so it is not covering the new behavior yet. Two reasons: `startTracesSdk` logs the same `Traces SDK won't start` message on its own, and this test does not set a root `batchProcessorConfig`, so the logs signal uses the 1000ms default while the test waits 15ms. That makes `expect(fetchSpy).not.toHaveBeenCalled()` pass even when the logs provider did start and queued an export. The other tests in this file pass the short delay. With the suggestion below, deleting the loop makes this test fail like it should.

```suggestion
    browserSdk = startBrowserSdk({
      // NOTE: short delay so a logs signal that wrongly started would flush
      // within the wait below, which is what `fetchSpy` asserts against.
      batchProcessorConfig: {
        scheduledDelayMillis: SCHEDULE_DELAY,
      },
```

packages/sdk/src/startBrowserSdk.test.ts
```diff
@@ -103,6 +104,36 @@ describe('startBrowserSdk', () => {
...
+        exportConfig: { url: 'this_is_not_an_URL' },
+      },
+    });
+    logs.getLogger('logs-sdk-test').emit({ eventName: 'test' });
+    trace.getTracer('traces-sdk-test').startSpan('test').end();
+    await new Promise((r) => setTimeout(r, SCHEDULE_DELAY + 5));
+
+    // Assert
+    expect(diagErrorSpy).toHaveBeenCalled();
+    expect(diagErrorSpy.mock.lastCall?.[0]).toMatch(/Traces SDK won't start/);
```
overbalance: This regex also matches `No Span processors configured. Traces SDK won't start`, so the test would still pass if the URL check went away and the bail-out came from somewhere else. I think we should match the URL message instead

```suggestion
    expect(diagErrorSpy.mock.lastCall?.[0]).toMatch(
      /Invalid OTLP export URL "this_is_not_an_URL". Traces SDK won't start/,
    );
```

Related, and outside the diff so I cannot suggest it inline: `diagErrorSpy` is created at line 50 but the `afterEach` at line 61 only clears `fetchSpy`, so it keeps calls from earlier tests. The test just above logs the same `Invalid OTLP export URL` string, which means `expect(diagErrorSpy).toHaveBeenCalled()` on the line above is already satisfied before this test even runs. Adding `diagErrorSpy.mockClear();` next to `fetchSpy.mockClear();` would fix it. The logs and traces suites have the same gap.

sandbox/src/otel.ts
```diff
@@ -69,16 +69,6 @@ export async function initOtel(
   customAttrs: Record<string, string> = {},
   { onSpan, onLog }: InitOtelOptions = {},
 ): Promise<OtelHandle> {
-  // ── Validate export endpoints ───────────────────────────────────────────────
```
overbalance, line 72: With this guard gone the sandbox no longer shows the failure. `startBrowserSdk` is called at line 116 without keeping the return value, so `initOtel` always resolves, `App.tsx` takes the `.then` branch and shows `SDK ready · <the invalid url>`. The `.catch` that used to show `SDK init failed` cannot be reached for this case any more.

It is also a little worse than before, because the no-op SDK throws away the sandbox's own session processors and console exporters too, and `registerInstrumentations` then wires everything to no-op providers. So the sandbox goes completely inert rather than just losing OTLP export. Since `tracesUrl` comes from the query string this is easy to hit by accident.

Could we keep the guard until `startBrowserSdk` gives callers something to check? Right now `NOOP_SDK` and a real SDK are the same shape, so there is no way for `initOtel` to tell them apart.

packages/sdk/README.md
```diff
@@ -275,6 +275,9 @@ Object containing configuraiton options for the HTTP log record exporter. These
 Note: you can pass this option to `startBrowserSdk` if you want to apply the same to all signals. If
 the option is defined at the top level and within the signal configuration the later wins.
 
+Note: an invalid `url` stops the SDK from starting — it logs a `diag.error` and returns a no-op SDK
+instead of starting and silently dropping the telemetry it cannot export.
```
overbalance: Could we say that the whole SDK stops here, not just this signal? An invalid logs URL also stops traces, and the `processors` the healthy signal defined get dropped as well. Saying the URL must be absolute http(s) would help too, given the `localhost:4318` case.

```suggestion
Note: an invalid `url` stops the whole browser SDK from starting, including the other signals and any
`processors` they define. It logs a `diag.error` and returns a no-op SDK instead of starting and
silently dropping the telemetry it cannot export. The `url` must be an absolute `http` or `https`
URL.
```

The same note is at line 317 for traces. And while you are in this file: the `processors` sections at lines 290 and 333, and the matching docs in `core/types.ts`, still say that setting `processors` makes the SDK ignore `exportConfig`. The code does not do that. `startLogsSdk` uses `if (!config?.processors || config?.exportConfig)`, so `exportConfig` is still honored, and the three new tests in this PR depend on it. The comment at `sdk.ts:129-131` repeats the same claim. Worth correcting so the new behaviour is not documented as impossible?

packages/sdk/src/startBrowserSdk.test.ts
mquentin: Wow thanks. I saw this in another test but completely forgot about adding it.

sandbox/src/otel.ts
mquentin: Very good point, I will introduce a new:

```
const INVALID_CONFIG_SDK: WebSdk = {
  invalidConfig: true,
  shutdown: () => Promise.resolve(),
};
```

On top of the existing `NOOP_SDK`.

This new `INVALID_CONFIG_SDK` will then throw an error in `sandbox/src/otel.ts`.

I think it makes sense and is also feature proof if we want a stronger configuration validation at one point

mquentin:
So @overbalance to address both of you feedbacks I introduced two commits:

- https://github.com/open-telemetry/opentelemetry-browser/pull/357/commits/d44983196a78f3dab19840184362556dfaedcfbf - that address the doc miss alignment
- https://github.com/open-telemetry/opentelemetry-browser/pull/357/commits/b30514b5691ea5950ff427810941724aec757d15 - that make sure we can now differentiate the NOOP_SDKs from the invalid SDKs and then, later like in the [sandbox example prevent registering any instrumentations](https://github.com/open-telemetry/opentelemetry-browser/pull/357/changes#diff-509798cb64a33272f60a5dd2e4f173a44a585e858f9be30909354ffa0562359fR143)

david-luna:
@mquentin could you fix the merge conflict?

mquentin:
> @mquentin could you fix the merge conflict?

@david-luna sorry for the delay, here is the fix pushed

david-luna (approved):

packages/sdk/src/core/types.ts
```diff
@@ -199,15 +202,26 @@ export type TracesConfig = CommonConfig & {
...
 
 export interface WebSdk {
+  /**
+   * `true` when the SDK refused to start because its configuration is invalid
+   * — e.g. a malformed export URL or no usable processors. Left `undefined` on
+   * a started SDK and on one intentionally turned off via `config.disabled`
+   * (that is not an error). Callers can check this to surface a
+   * misconfiguration, e.g. `if (sdk.invalidConfig) { throw ... }`.
+   */
+  invalidConfig?: boolean;
```
joaquin-diaz, line 225: Do we think that in the future we might have something else triggering a failure in SDK start that is not config related?  I can't think of an example right now but I'm wondering if we would cage ourselves with this name and force us to change it in the future or add another flag for a different kind of start up failure

Maybe we can rename it something more generic like `startFailed` or just `failed`

david-luna: If we rename the property (to failed as an example) I think we should change the type so it can hold the reason.

joaquin-diaz (approved):

PR #358, opened 2026-07-20
feat(user-action): add callback function to create custom attributes in user action logs

ssakhamu:
## Which problem is this PR solving?

In other instrumentation packages (navigation, web-vitals, etc), the applyCustomLogRecordData hook allows custom attributes to be attached to logs. This hook does not exist in the user-action instrumentation; this change adds this hook into the user-action instrumentation with the appropriate tests. The goal is to provide consistency in adding custom attributes to any logs generated across instrumentation packages.

## Short description of the changes

- adds the applyCustomLogRecordData hook to UserActionInstrumentationConfig
- safely executes this hook on the generated log record in the event of a click
- tests the success and failure cases of the hook

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] if the applyCustomLogRecordData hook is present in the config passed into UserActionInstrumentation, ensure that the custom attributes it creates are in the log record
- [x] if the applyCustomLogRecordData hook throws an error, ensure that the user action log is still emitted and the error is logged via diagnostics logger

These two tests are in packages/instrumentation/src/user-action/instrumentation.test.ts.

## Checklist:

- [x] Followed the style guidelines of this project
- [x] Unit tests have been added
- [x] Documentation has been updated

Issue #359, opened 2026-07-22
Consolidate applyCustomLogRecordData into a single SDK-level log configuration


martinkuba:
### Problem

Three instrumentations now expose an identical `applyCustomLogRecordData` hook (navigation, web-vitals, [user-action](https://github.com/open-telemetry/opentelemetry-browser/pull/358)), each with near-duplicate implementation. A user who wants to attach the same attributes to all telemetry has to configure every instrumentation individually and we we also end up with duplicated code across multiple instrumentations.

There is no single place to enrich every log record.

### Proposal

Add a single config setting at the SDK level that the SDK wraps in a LogRecordProcessor. 

Additionally, we could deprecate the per-instrumentation `applyCustomLogRecordData` hooks in favor of the global option. Users targeting a specific instrumentation can use conditional logic on `logRecord.eventName`.

Keeping both means two overlapping mechanisms with different signatures (SdkLogRecord.setAttribute vs API LogRecord.attributes), which is confusing.

### Decisions to make

- Custom logs - processor would also enrich logs the user emits directly. I think this is what we want, but should confirm.
- Spans - should the same pattern extend to spans?
- Deprecation path - deprecate existing per-instrumentation callbacks or keep both?

### Out of scope

PR #358 adds the hook to user-action instrumentation, which is consistent with the existing pattern. This issue is the forward-looking consolidation, not a blocker for that PR.

ayishaatwork:
Hi! I'd like to work on this issue if it's still available.

I'll first review the existing `applyCustomLogRecordData` implementations across the instrumentations, understand the current logging pipeline, and explore how this could be consolidated at the SDK level. I'll share the proposed approach if any design questions need to be resolved before opening a PR.

Please let me know if there are any preferences or prior discussions I should follow before getting started.

david-luna:
Something I've noticed is the function type does not expect accept any other param than the log record. Today I got a question on how to add customs attributes based on the element (or parent) clicked and IMHO the log record provides limited information (tag name, selector). I think it would be good to have the event that originated the log.

You can see a similar patter in fetch and XHR instrumentations with `applyCustomAttributesOnSpan` where the function gets the Span and the thing that originated it (Request or XHR object). For logs would be something like.

```ts
export type ApplyCustomLogRecordDataFunction = (logRecord: LogRecord, event: Event | undefined) => void;
```

[...]

PR #360, opened 2026-07-24
feat(navigation-timing): add callback function to create custom attributes in navigation timing logs

ssakhamu:

In other instrumentation packages (navigation, web-vitals, user-action, etc), the applyCustomLogRecordData hook allows custom attributes to be attached to logs. This hook does not exist in the navigation-timing instrumentation; this change adds this hook into the navigation-timing instrumentation with the appropriate tests. The goal is to provide consistency in adding custom attributes to any logs generated across instrumentation packages. This is a mirrored change to #358.

- adds the applyCustomLogRecordData hook to NavigationTimingInstrumentationConfig
- safely executes this hook on the generated log record in a navigation timing event
- tests the success and failure cases of the hook

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- [x] if the applyCustomLogRecordData hook is present in the config passed into NavigationTimingInstrumentation, ensure that the custom attributes it creates are in the log record
- [x] if the applyCustomLogRecordData hook throws an error, ensure that the navigation timing log is still emitted and the error is logged via diagnostics logger

These two tests are in packages/instrumentation/src/navigation-timing/instrumentation.test.ts.

## Checklist:


packages/instrumentation/src/navigation-timing/instrumentation.test.ts
```diff
@@ -542,4 +542,62 @@ describe('NavigationTimingInstrumentation', () => {
...
+
+      getEntriesByTypeSpy.mockReturnValueOnce([entry]);
+
+      expect(() => {
+        instrumentation.enable();
+      }).not.toThrow();
+
+      const logs = getNavigationTimingLogs();
+      expect(logs.length).toBe(1);
+    });
```
overbalance, line 613: This passes even with the `this._diag.error('applyCustomLogRecordData hook failed', error)` line deleted from the instrumentation, so the reporting is not covered. It is not visible in the output either: `setupTestLogExporter` does not install a diag logger, so diag does nothing during tests and a spy is the only way to check it.

This is the same spy that `navigation/instrumentation.test.ts` and `user-action/instrumentation.test.ts` already use.

```suggestion
    it('should catch errors thrown by the hook and still emit', () => {
      instrumentation = new NavigationTimingInstrumentation({
        enabled: false,
        applyCustomLogRecordData: () => {
          throw new Error('hook error');
        },
      });

      const diagErrorSpy = vi
        .spyOn(
          (
            instrumentation as unknown as {
              _diag: { error: (...args: unknown[]) => void };
            }
          )._diag,
          'error',
        )
        .mockImplementation(() => {});

      const entry = {
        name: 'https://example.test/',
        entryType: 'navigation',
        startTime: 0,
        duration: 123,
        type: 'navigate',
        loadEventEnd: 456,
      };

      getEntriesByTypeSpy.mockReturnValueOnce([entry]);

      expect(() => {
        instrumentation.enable();
      }).not.toThrow();

      const logs = getNavigationTimingLogs();
      expect(logs.length).toBe(1);
      expect(diagErrorSpy).toHaveBeenCalled();
    });
```

packages/instrumentation/src/navigation-timing/instrumentation.test.ts
```diff
@@ -542,4 +542,62 @@ describe('NavigationTimingInstrumentation', () => {
 
     addEventListenerSpy.mockRestore();
   });
+
+  describe('applyCustomLogRecordData hook', () => {
+    it('should allow hook to add custom attributes', () => {
+      const instrumentation = new NavigationTimingInstrumentation({
```
overbalance: `const` here creates a new local variable that hides the shared `instrumentation` from the top of the `describe` block, so the `disable()` in `afterEach` runs on the other instance and this one is never disabled. It does not break anything today, only because `loadEventEnd: 456` makes the record emit immediately and clean up after itself. The test below already assigns to the shared variable.

```suggestion
      instrumentation = new NavigationTimingInstrumentation({
```

packages/instrumentation/src/navigation-timing/instrumentation.ts
```diff
@@ -232,7 +236,26 @@ export class NavigationTimingInstrumentation extends InstrumentationBase<Navigat
...
       },
-    });
+    };
+
+    this._applyCustomLogRecordData(logRecord);
+
+    this.logger.emit(logRecord);
+  }
+
+  private _applyCustomLogRecordData(logRecord: LogRecord) {
```
overbalance:
```suggestion
  private _applyCustomLogRecordData(logRecord: LogRecord): void {
```

overbalance:
Nice work on this, thanks for picking it up 🙌

The implementation looks right and matches the user-action version from #358. Every emit path goes through `_emitNavigationTiming`, so the hook cannot be skipped.

I left three small suggestions on the tests, all committable straight from the review:

1. the error test does not check that the failure is reported (deleting the `_diag.error(...)` line keeps it green)
2. `const instrumentation` in the first test hides the shared variable, so `afterEach` disables the wrong instance
3. a return type on the new private method

Nothing else needed from you. The other points I noticed apply equally to the three instrumentations that already have this hook, so they are not from your change and I will raise them separately.

packages/instrumentation/src/navigation-timing/instrumentation.test.ts
ssakhamu: added diagErrorSpy

packages/instrumentation/src/navigation-timing/instrumentation.test.ts
ssakhamu: added

packages/instrumentation/src/navigation-timing/instrumentation.ts
ssakhamu: added

martinkuba:
If the function does not get any additional context specific to this instrumentation, then I don't think this provides any value. See [this comment](https://github.com/open-telemetry/opentelemetry-browser/issues/359#issuecomment-5282391676) for more context.

If there is a real use case for adding a custom attribute based on context from this instrumentation, then let's discuss further. Otherwise, I would be inclined not to add.

martinkuba:
Per the decision discussed in https://github.com/open-telemetry/opentelemetry-browser/issues/359, I am closing this since there is no additional context provided to the callback. @ssakhamu If you disagree, please feel free to reopen this.

PR #361, opened 2026-07-24
docs: add roadmap doc

martinkuba:
Per the SIGs agreement, I am adding a roadmap doc that describes the areas we want to put our focus in going forward.

This is a starting point for a discussion. Please comment if there is anything you think is inaccurate or missing.

I have used the [Android SIG roadmap](https://github.com/open-telemetry/opentelemetry-android/blob/main/docs/ROADMAP.md) as template - copying the format was intentional.

/cc @overbalance @joaquin-diaz @wolfgangcodes @david-luna


docs/roadmap.md
```diff
@@ -0,0 +1,276 @@
...
+
+Sessions and page context are central to browser observability and a key
+differentiator from server-side telemetry. Goals:
+
+* First-class session management in the SDK, including session start/end
+  events.
+* Modeling sessions and documents/page views as entities, with session and
+  page context carried on telemetry via resource attributes
+  ([#269](https://github.com/open-telemetry/opentelemetry-browser/pull/269),
+  [semantic-conventions#3519](https://github.com/open-telemetry/semantic-conventions/pull/3519)).
```
mquentin: Super small nit:

this link is the one of the closed PR, the one that was merged is:
https://github.com/open-telemetry/semantic-conventions/pull/3633

mquentin (approved):
Nice addition ! I added a super small nit

docs/roadmap.md
```diff
@@ -0,0 +1,276 @@
...
+* [Documentation](#documentation)
+* [Cross-Platform Coordination](#cross-platform-coordination)
+* [Growing Contributors](#growing-contributors)
+* [Distant Destinations](#distant-destinations)
+  * [Bundle Size](#bundle-size)
+  * [Protocol and Payload Efficiency](#protocol-and-payload-efficiency)
+  * [Sampling](#sampling)
+* [Conclusion](#conclusion)
+
+## Background: From Phase 1 to This Roadmap
```
rebehe, line 41: Would it be helpful to document some key principles to help explain why certain decisions were made up to this point? e.g.:

- **Keep client bundle sizes small**: Clients care a lot about preserving limited client-side resources. Anything that creates more computational burden when it isn’t necessary - e.g. keeping a span open for longer than it needs to, using a span when it could be a log - is discouraged. Let clients focus on emitting high fidelity events and defer any aggregation or computation to the backend.
- **Don’t create client-side spans with ambiguous start or end times**: Client-side interactions are highly asynchronous, so it can be ambiguous to determine when a client-side span should end. In cases like this, use logs instead.
- **Be pragmatic when dealing with existing browser constraints**: Browsers have a few interesting limitations that can lead to complex telemetry design:
    1. **_Browsers lack native context propagation_**: There is no native browser solution for maintaining context from a user click to an asynchronous promise (e.g. a network request). Current workarounds, like Zone.js, are clunky and intrusive. Because there’s no good way to solve the context propagation gap, user interactions will be captured as point-in-time events, and instead rely on timestamps for implicit user interaction/network request association.
    1. **_W3C browser timing/performance APIs are asynchronous_**: These APIs capture detailed sub-millisecond durations for HTTP and page load milestones. They also populate asynchronously with a slight lag, so if a user navigates away from the page, it could result in this data being dropped. Instead, if this timing information is captured asynchronously as logs, this data will be emitted even if the user navigates away from the page, and the network or page load span is allowed to end at the precise time of page network response or page load completion.

I know this thinking is mentioned a few times across various docs, but having it explicitly written down could be helpful. I also recognize that the roadmap doc might not be the right place for it, so feel free to point me somewhere else.

martinkuba: I agree that this would be useful. It could go in the [browser-observability-model](https://github.com/open-telemetry/opentelemetry-browser/blob/main/docs/browser-observability-model.md) doc, or we could add a separate `design-principles` doc (which the model page could link to).

Would you be up for opening a PR? Or I can create an issue and start a draft.

docs/roadmap.md
martinkuba: I have updated the link

docs/roadmap.md
rebehe: I'll take a look at the browser-observability-model doc 👍 and maybe poke around at the other docs too to see what might make the most sense.

This is also aligned with what we talked about yesterday in the SIG around onboarding. I'll take the action item to open up a PR (let me know if I should also create an issue too, or if that's not necessary).

docs/roadmap.md
```diff
@@ -0,0 +1,276 @@
...
+to minimize churn for users. As part of this consolidation, we are committed
+to a zoneless architecture
+([#210](https://github.com/open-telemetry/opentelemetry-browser/issues/210)):
+new packages will not depend on Zone.js, and the legacy span-based
+user-interaction instrumentation will be deprecated in favor of event-based
+user action telemetry.
+
+## Event-Based Instrumentation
+
+Browser telemetry in this repository is built on an event-based model,
```
joaquin-diaz: Does `event-based` include spans?

martinkuba: I have renamed the section and also removed the paragraph about instrumentations being event-based. There is no reason to have this distinction described in this doc.

docs/roadmap.md
martinkuba: @rebehe Thanks for taking this on. No need for an issue if you are planning to open a PR yourself.

rebehe: https://github.com/open-telemetry/opentelemetry-browser/pull/380

wolfgangcodes (approved):

david-luna (approved):

PR #367, opened 2026-07-28
feat(browser-instrumentation): add XHR instrumentation

### LINKED ISSUE #258: Move XHR instrumentation from JS core to this repo
Sub-issue of https://github.com/open-telemetry/opentelemetry-browser/issues/209

Move the `@opentelemetry/instrumentation-xml-http-request` instrumentation from [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-xml-http-request).

Related to  https://github.com/open-telemetry/opentelemetry-js/issues/6591 and https://github.com/open-telemetry/opentelemetry-js/pull/6630

Decisions needed
- should it be part of the consolidated instrumentation package or separate? (the current package is widely used)
- where should it import utils from (can be separate issue)
- should we keep the span events for resource timing or make it work with the new resource timing instrumentation (can be separate issue)

david-luna:

Closes #258

Inclusion of the `XmlHttpRequest `instrumentation with some differences from the original version in https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-fetch

- utils moved to this package instead of using `@opentelemetry/sdk-trace-web`
- removed dependency from `@opentelemetry/core`
- only emit span for the XHR call. it does not emit child spans for preflight requests neither span event for the resource timings

this instrumentation follows the previous work on `fetch` instrumentation ([PR](https://github.com/open-telemetry/opentelemetry-browser/pull/281)) by relying on the `resource-timing` instrumentation to correlate logs.

## Type of change

## How Has This Been Tested?

- [X] Unit tests for new utils
- [X] Browser tests for instrumentation

## Checklist:


packages/instrumentation/src/xhr/types.ts
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { Span } from '@opentelemetry/api';
+import type { InstrumentationConfig } from '@opentelemetry/instrumentation';
+
+type XhrCustomAttributeFunction = (
+  span: Span,
+  xhr: XMLHttpRequest,
+  // XXX: fetch instrumentation has a result paranm here (should we align?)
```
david-luna, line 12: open to discussion but I think it wont be easy to have a response object here.

packages/instrumentation/src/xhr/types.ts
```diff
@@ -0,0 +1,33 @@
...
+) => void;
+
+// XXX: this configuration is very similar to fetch instrumentation config.
```
david-luna, line 15: I think the only property with different type is `applyCustomAttributesOnSpan` and also this instrumentation does not have the `requestHook`. Should we have an equivalent hook ?

packages/instrumentation/src/xhr/instrumentation.test.ts
```diff
@@ -0,0 +1,933 @@
...
+      const startTime = performance.now();
+      await doXhrRequest({ method: 'GET', url }).then((r) => r.json());
+      const endTime = performance.now();
+
+      // Span is exported
+      const span = await waitForSpan(url);
+      expect(span.name).toBe('GET');
+      expect(span.kind).toEqual(SpanKind.CLIENT);
+      expect(span.attributes[ATTR_HTTP_REQUEST_METHOD]).toEqual('GET');
+      expect(span.attributes[ATTR_URL_FULL]).toEqual(url);
```
joaquin-diaz: In this case we're creating a span with `ATTR_URL_FULL = '/api/get'`. That breaks the convention given that is not actually the full url, is a relative url. I think that wouldn't then match with the resource timing entry later given that we match using [PerformanceResourceTiming.name](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry/name) that maps to the absolute URL.

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,307 @@
...
+  private _endSpan(xhr: XMLHttpRequest, errorType?: string) {
+    const spanDetails = this._xhrSpanMap.get(xhr);
+
+    if (spanDetails) {
+      const { span, url, start } = spanDetails;
+
+      span.setAttribute(ATTR_HTTP_RESPONSE_STATUS_CODE, xhr.status);
+      // https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status
+      const isErrorStatus = xhr.status < 200 || xhr.status >= 400;
+      if (isErrorStatus || errorType) {
```
joaquin-diaz: This would mark an aborted request with `xhr.status = 0` as error given that `xhr.status < 200 == true`. According to the conventions this should be unset

> If the HTTP client instrumentation can detect that the request was cancelled intentionally by the caller (e.g., via context cancellation or an abort signal), the cancellation SHOULD NOT be treated as an error: the span status SHOULD be left unset and error.type SHOULD NOT be set.

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,307 @@
...
+    return this.tracer.startSpan(normMethod, {
+      kind: SpanKind.CLIENT,
+      attributes,
+    });
+  }
+
+  /**
+   * Finish span, add attributes, network events etc.
+   */
+  private _endSpan(xhr: XMLHttpRequest, errorType?: string) {
```
joaquin-diaz: Can we type errorType? Maybe using an enum or something that makes us not having hardcoded values like "Request error" or "Request timeout"

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,307 @@
...
+      attributes[ATTR_HTTP_REQUEST_METHOD_ORIGINAL] = origMethod;
+    }
+    attributes[ATTR_URL_FULL] = sanitizeUrl ? sanitizeUrl(url) : url;
+    attributes[ATTR_SERVER_ADDRESS] = parsedUrl.hostname;
+    const serverPort = serverPortFromUrl(parsedUrl);
+    if (serverPort) {
+      attributes[ATTR_SERVER_PORT] = serverPort;
+    }
+
+    return this.tracer.startSpan(normMethod, {
```
joaquin-diaz, line 239: This starts the span right away, even though a user can create an xhr request, open it and send it at any point:

```
const xhr = new XMLHttpRequest();   
xhr.open('GET', '/api/users');     
// User can do anything here, maybe an async task that takes a long time
xhr.send();                         
```

The gap between `open` and `send` is unknown for us and it can inflate the overall span duration with extra time that is not necessary related to the network request itself. The old instrumentation does the same, but it has span events for when `send` is call so you could fix the inflated value but it still would be annoying instead of just looking at span duration directly. 

I think we should fix this in the new instrumentation but we can check with the sig, and also I don't think we need to add the span events back

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  /**
+   * Patches the constructor of fetch
+   */
+  private _patchSend(): (original: XhrSendFunction) => XhrSendFunction {
+    return (original) => {
+      const instrumentation = this;
+
+      return function patchedSend(
```
joaquin-diaz, line 162: The old instrumentation accounted for someone calling open/send multiple times for the same xhr object by removing all existing listeners and the entry from the weak map

https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L361

joaquin-diaz (commented):
I found a few inconsistencies between the implementation and the http span semconv, and It also seems that we missed them during the fetch instrumentation review so I created a PR [here](https://github.com/open-telemetry/opentelemetry-browser/pull/392)
with the fixes. I also added fetch to the e2e tests so we can validate it works with the resource timing instrumentation. Can you add e2e tests for this one?

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: There is no way to know if the cycles used between `open` and `send` calls are part of the request task (therefore is part of the Span) or not.

> and also I don't think we need to add the span events back


I agree. I my mental model a span event is just a correlated event/log. Maybe we could model it with a couple of logs with the right span context. Food for thought.

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: right, I think the logic could be almost identical to the fetch

packages/instrumentation/src/xhr/instrumentation.test.ts
david-luna: good catch, updated the implementation and the test

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: I've aligned the tests with the fetch ones and updated the error handling

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: used a regular type

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: updated the implementation to start the span when `send` is called

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: now the listeners are removed when the span ends

e2e-tests/packages/instrumentation/xhr/instrumentation.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    const span = await vi.waitFor(() => getSpanByUrl(url), { timeout: 2000 });
+    const log = await vi.waitFor(() => getResourceTimingLogByUrl(url), {
+      timeout: 5000,
+    });
+
+    expect(log.traceId).toBe(span.traceId);
+    expect(log.spanId).toBe(span.spanId);
+  });
+
+  it.only('correlates the resource timing entry when XHR is called with a relative URL', async () => {
```
joaquin-diaz: Remember to clear this. We should have a linter rule that prevents `only` in tests

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,337 @@
...
+    if (shouldPropagate) {
+      propagation.inject(ctx, xhr, {
+        set: (x, k, v) => {
+          x.setRequestHeader(k, typeof v === 'string' ? v : String(v));
+        },
+      });
+    } else {
+      const headers: Partial<Record<string, unknown>> = {};
+      propagation.inject(ctx, headers);
+      propagation.inject(ctx, headers, {
```
joaquin-diaz, line 336: Duplicated inject?

joaquin-diaz (commented):
Looking good! I think we're almost there. If the two TODOs that are left are not a regression vs the old instrumentation I think it would be ok to do them as a follow up.

Can you update `packages/instrumentation/README.md` adding the new instrumentation?

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,337 @@
...
+      // `_wrap` throws if a third-party script has locked globalThis.fetch via
+      // Object.defineProperty(XMLHttpRequest.prototype, 'open', { writable: false, ... }).
+      this._wrap(XMLHttpRequest.prototype, 'open', this._patchOpen());
+      this._wrap(XMLHttpRequest.prototype, 'send', this._patchSend());
+      this._isXhrPatched = true;
+      this._isEnabled = true;
+    } catch (err) {
+      // make sure there is no wrapped functions
+      this._unwrap(XMLHttpRequest.prototype, 'open');
+      this._unwrap(XMLHttpRequest.prototype, 'send');
```
overbalance: On the web we never want to unwrap, only disable, because other vendors may also have wrapped xhr and restoring it to the original would remove their functionality

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,337 @@
...
+        (error) => {
+          if (!error) {
+            return;
+          }
+          this._diag.error('applyCustomAttributesOnSpan', error);
+        },
+        true,
+      );
+    }
+  }
```
overbalance, line 317: @martinkuba made a good point in another PR that this responsibility is better placed in a Processor. what do you think?

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: Forgot to remove the 1st one. Inject needs a setter function to work properly

e2e-tests/packages/instrumentation/xhr/instrumentation.test.ts
david-luna: > We should have a linter rule that prevents only in tests


I agree.

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: I think is more straight forward for the current instrumentations that emit logs. The `applyCustomLogRecordData` function only gets the log with no context. In this case (like in fetch) the function receives the XHR object (fetch gets the response or an error) so user can inspect it and add attributes related to the info the object holds.

packages/instrumentation/src/xhr/instrumentation.ts
david-luna: I agree the handling copied from https://github.com/open-telemetry/opentelemetry-js/blob/8f103777b707029e9bb18b2755531771de1b80f5/experimental/packages/opentelemetry-instrumentation-xml-http-request/src/xhr.ts#L587 is not the best.

I think this one is tricky because we are patching 2 methods of the prototype. If `open` can be wrapped but not  `send` and try to enable it several times we may end up with a wrap of a wrap of a wrap...

Let's remove it for now and I'll run some tests on that scenario.

joaquin-diaz (approved):

overbalance (commented):
We spoke offline in the SIG meeting about using a cascading try/catch that reverts patching if either function fails. This should be safe because of its synchronous nature.

david-luna:
> We spoke offline in the SIG meeting about using a cascading try/catch that reverts patching if either function fails. This should be safe because of its synchronous nature.

commit https://github.com/open-telemetry/opentelemetry-browser/pull/367/commits/8aa869d598e06914c5cd053fc3ce731bbe055ef1

packages/instrumentation/src/xhr/instrumentation.ts
```diff
@@ -0,0 +1,345 @@
...
+    // Object.defineProperty(XMLHttpRequest.prototype, 'open', { writable: false, ... }).
+    try {
+      this._wrap(XMLHttpRequest.prototype, 'open', this._patchOpen());
+    } catch (err) {
+      this._diag.warn(
+        'Failed to patch XMLHttpRequest.prototype.open; instrumentation will not be enabled. ' +
+          'Another script may have locked XMLHttpRequest.prototype.open via Object.defineProperty.',
+        err,
+      );
+    }
```
overbalance, line 92: I suggest bailing right here and marking as "attempted". what do you think?
```suggestion
      );
      return;
    }
```

overbalance (approved):

PR #368, opened 2026-07-29
fix(browser-instrumentation): add max capacity to context registry

david-luna:

Refs: https://github.com/open-telemetry/opentelemetry-browser/pull/281/changes#r3672601773

Changes to avoid possible excess of memory allocation. The current implementation makes possible for the registry to keep growing in entries if there is more calls to `register` than `unregister. With these changes we keep a limit of entries in the registry. It also relaxes the `TData` type by adding a new method to resolve the key from the data object.

- add `getLookupKey` method
- add `getDataKey` method
- loosen the types on ContextRegistry by changing the abstract methods. `TData` does not need a `key` field since is resolved by `getDataKey`
- add `MAX_CAPACITY` to the registered entries. The registry removes the oldest item when capacity is reached and new data is registered.

## Type of change

## How Has This Been Tested?

- [X] run unit tests

## Checklist:

- [X] Unit tests have been updated


packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,29 +18,45 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
  * @typeParam TLookup - The object the consuming instrumentation passes to
  *   `getContext` and `unregister` (e.g. a `PerformanceResourceTiming` entry).
  */
-export abstract class ContextRegistry<TData extends { key: string }, TLookup> {
+export abstract class ContextRegistry<TData, TLookup> {
   protected _records = new Map<string, StoredRecord<TData>[]>();
+  private _usedKeys: string[] = [];
 
-  /** Store the span context for the given data, indexed by `data.key`. */
+  /** Store the span context for the given data, key is resolved . */
```
overbalance:
```suggestion
  /** Store the span context for the given data, key is resolved by getDataKey(). */
```

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,29 +18,45 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
     const ctx = trace.setSpan(ROOT_CONTEXT, span);
-    const list = this._records.get(data.key) ?? [];
+    const key = this.getDataKey(data);
+    const list = this._records.get(key) ?? [];
     list.push({ ...data, ctx });
-    this._records.set(data.key, list);
+    this._records.set(key, list);
+
+    // To keep track of the items added we just need the keys. Removing the oldest
+    // bocomes only keeping the key used at that time and remove the 1st element of
```
overbalance:
```suggestion
    // becomes only keeping the key used at that time and remove the 1st element of
```

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,29 +18,45 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
+    // the list under that key (since we are pushing)
+
+    // Keep the key used for registration
+    this._usedKeys.push(key);
```
overbalance: I think `unregister()` never removes anything from `_usedKeys`, so the two get out of sync. After about 1000 requests to a repeated URL, the key we shift off is the same one we just pushed, so `oldestList.shift()` deletes the record added two lines above. I tried it with 5 rotating URLs and every registration after the 1000th was lost. Could we store the record in `_usedKeys` instead of just the key, and remove it in `unregister()` too?

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,29 +18,45 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
+    if (this._usedKeys.length > MAX_CAPACITY) {
+      // 1st key of the array is the oldest
+      const oldestKey = this._usedKeys.shift() as string;
+      // and 1st item on the list is the oldest
+      const oldestList = this._records.get(oldestKey) || [];
+      oldestList.shift();
```
overbalance: `oldestList.shift()` empties the list but leaves the key in `_records`, so with cache-busted URLs the map still grows forever. 3000 registrations leaves 3000 keys, 2000 of them empty arrays. Could we delete the key when the list is empty, like `unregister()` already does at line 71?

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,29 +18,45 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
+
+    // To keep track of the items added we just need the keys. Removing the oldest
+    // bocomes only keeping the key used at that time and remove the 1st element of
+    // the list under that key (since we are pushing)
+
+    // Keep the key used for registration
+    this._usedKeys.push(key);
+    if (this._usedKeys.length > MAX_CAPACITY) {
+      // 1st key of the array is the oldest
+      const oldestKey = this._usedKeys.shift() as string;
```
overbalance: Small one, could we do `if (oldestKey === undefined) return;` instead of the cast? It is safe today because of the length check above, but the cast would hide it if that check ever changes.

packages/instrumentation/src/utils/ContextRegistry.test.ts
```diff
@@ -14,7 +14,11 @@ interface TestData {
 }
 
 class TestRegistry extends ContextRegistry<TestData, string> {
-  getKey(lookup: string): string {
+  getDataKey(data: TestData): string {
```
overbalance, line 17: Could we add a test for the eviction? At the moment the suite still passes if the whole `if` block in `register()` is removed, so nothing covers the new behaviour.

overbalance (changes requested):

packages/instrumentation/src/utils/ContextRegistry.ts
david-luna: > Could we store the record in _usedKeys instead of just the key, and remove it in unregister() too?

I used keys on purpose to simplify management. I think the initial implementation using a Map of arrays was intended for performance and having to manage an array with all the records goes against it.

Maybe keeping a single record list would be enough if the capacity is not too big?

david-luna: I check & the lookup operation keep being performant so I guess it's okay to keep another array around. I think other comments no longer apply if we change the data structure.

commit https://github.com/open-telemetry/opentelemetry-browser/pull/368/commits/e73d492c226547cf231757c6beaa8647a6bb9af4

packages/instrumentation/src/utils/ContextRegistry.test.ts
david-luna: commit https://github.com/open-telemetry/opentelemetry-browser/pull/368/commits/e73d492c226547cf231757c6beaa8647a6bb9af4

david-luna:
@overbalance I did use a record array as you suggested. Thanks for the feedback :)

packages/instrumentation/src/utils/ContextRegistry.ts
```diff
@@ -16,49 +18,92 @@ type StoredRecord<TData> = TData & { ctx: Context };
...
   unregister(lookup: TLookup): void {
-    const key = this.getKey(lookup);
+    const key = this.getLookupKey(lookup);
     const ctx = this.getContext(lookup);
     if (ctx === undefined) {
       return;
     }
 
-    const list = this._records.get(key);
+    // Remove form the lookup map and from the global list
```
overbalance:
```suggestion
    // Remove from the lookup map and from the global list
```

overbalance (approved):

PR #369, opened 2026-07-29
fix(sdk): accept sampler option in traces SDK

david-luna:

The sampler option is available in the traces SDK configuration but the implementation never used it to configure the tracer provider. The code was commented in in a `TODO` for a further discussion ([ref](https://github.com/open-telemetry/opentelemetry-browser/blob/533fcdf72db5eed6158b9e2bc42deb2452236d98/packages/sdk/src/traces/startTracesSdk.ts#L78-80)).

This PR adds the support for a sampler option and includes a test.

- pass sampler configuration to `TracerProvider` when calling the constructor.
- add a test to check the sampler is working.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [X] unit test

## Checklist:

### DIFF
--- a/packages/sdk/src/traces/startTracesSdk.test.ts
+++ b/packages/sdk/src/traces/startTracesSdk.test.ts
@@ -4,7 +4,10 @@
  */
 
 import { diag, trace } from '@opentelemetry/api';
-import { SimpleSpanProcessor } from '@opentelemetry/sdk-trace';
+import {
+  AlwaysOffSampler,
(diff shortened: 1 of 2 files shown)


mquentin (approved):
LGTM with unit test.

## nit for the future startTracesSdk API changes
it would be nice to add all these startTracesSdk config options in the sandbox as a form toggle to be able to QA on a real application that the config works as intended

PR #392, opened 2026-08-11
feat: add e2e tests for fetch instrumentation

joaquin-diaz:
## Summary
- Add e2e tests for `FetchInstrumentation`: span creation, non-2xx responses, network errors, aborts, parent/child context, and correlation with `ResourceTimingInstrumentation`
- Found and fixed three discrepancies of the [HTTP span status semantic convention](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status) in `FetchInstrumentation`:
  - Intentional request cancellation (`AbortController`) was treated as an error instead of leaving the span status unset with no `error.type`.
  - `http.response.status_code` was set to `0` for network errors/aborts instead of being omitted (no response was ever received).
  - `error.type` used the HTTP status reason phrase (e.g. `"Internal Server Error"`) instead of the numeric status code as a string (e.g. `"500"`), and the exception's `message` instead of its `name` for pre-response errors.

### DIFF
--- a/e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
+++ b/e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
@@ -1,0 +1,374 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { context, trace } from '@opentelemetry/api';
+import { FetchInstrumentation } from '@opentelemetry/browser-instrumentation/experimental/fetch';
+import { ResourceTimingInstrumentation } from '@opentelemetry/browser-instrumentation/experimental/resource-timing';
+import {
+  ATTR_ERROR_TYPE,
+  ATTR_HTTP_REQUEST_METHOD,
+  ATTR_HTTP_RESPONSE_STATUS_CODE,
+  ATTR_SERVER_ADDRESS,
+  ATTR_SERVER_PORT,
+  ATTR_URL_FULL,
+} from '@opentelemetry/semantic-conventions';
+import { HttpResponse, http } from 'msw';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { mockServer } from '../../../utils/mock-server.ts';
+import {
+  COLLECTOR_URL,
+  collector,
+  LOGS_COLLECTOR_URL,
+} from '../../../utils/test-collector.ts';
+import type { TestSdkHandle } from '../../../utils/test-otel-setup.ts';
+import { testSdkSetup } from '../../../utils/test-otel-setup.ts';
+
+// The SDK exports spans/logs over real HTTP via fetch(). Without excluding
+// the collector endpoints, FetchInstrumentation would create a span for each
+// export call, which would itself need exporting, recursing forever.
+const IGNORE_COLLECTOR_URLS = [COLLECTOR_URL, LOGS_COLLECTOR_URL];
+
+const getUrlForPath = (path: string) => new URL(path, location.href).href;
+
+// This test's own fake application endpoints, unrelated to the OTLP collector.
+const fetchHandlers = [
+  http.get('/e2e/fetch/get', () => HttpResponse.json({ ok: true })),
+  http.get(
+    '/e2e/fetch/error',
+    () =>
+      new HttpResponse(null, {
+        status: 500,
+        statusText: 'Internal Server Error',
+      }),
+  ),
+  http.get('/e2e/fetch/network-error', () => HttpResponse.error()),
+  http.get('/e2e/fetch/abort', () => HttpResponse.json({ ok: true })),
+  http.get('/e2e/fetch/child', () => HttpResponse.json({ ok: true })),
+  http.get('/e2e/fetch/resource-correlation', () =>
+    HttpResponse.json({ ok: true }),
+  ),
+  http.get('/e2e/fetch/relative-resource-correlation', () =>
+    HttpResponse.json({ ok: true }),
+  ),
+];
+
+describe('FetchInstrumentation', () => {
+  let result: TestSdkHandle;
+
+  beforeEach(() => {
+    mockServer.use(...fetchHandlers);
+  });
+
+  afterEach(async () => {
+    await result.shutdown();
+  });
+
+  it('creates a span for a fetch request', async () => {
+    result = testSdkSetup([
+      new FetchInstrumentation({ ignoreUrls: IGNORE_COLLECTOR_URLS }),
+    ]);
+
+    const url = getUrlForPath('/e2e/fetch/get');
+    await fetch(url);
+
+    await vi.waitFor(
+      () => {
+        const span = collector
+          .getSpans()
+          .find(
+            (s) =>
+              s.attributes.find((a) => a.key === ATTR_URL_FULL)?.value
+                .stringValue === url,
+          );
+        expect(span).toBeDefined();
+        if (!span) {
+          throw new Error('Span is undefined');
+        }
+
+        const attr = (key: string) =>
+          span.attributes.find((a) => a.key === key)?.value;
+
+        expect(attr(ATTR_HTTP_REQUEST_METHOD)).toEqual({
+          stringValue: 'GET',
+        });
+        expect(attr(ATTR_SERVER_ADDRESS)).toEqual({
+          stringValue: location.hostname,
+        });
+        expect(attr(ATTR_SERVER_PORT)).toBeDefined();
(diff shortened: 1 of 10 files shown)


e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
```diff
@@ -0,0 +1,374 @@
...
+import { mockServer } from '../../../utils/mock-server.ts';
+import {
+  COLLECTOR_URL,
+  collector,
+  LOGS_COLLECTOR_URL,
+} from '../../../utils/test-collector.ts';
+import type { TestSdkHandle } from '../../../utils/test-otel-setup.ts';
+import { testSdkSetup } from '../../../utils/test-otel-setup.ts';
+
+// The SDK exports spans/logs over real HTTP via fetch(). Without excluding
```
david-luna: For this issue https://github.com/open-telemetry/opentelemetry-js/pull/6353 access the `__original` property of fetch's wrapper function. Have you experienced the issue in these tests?

e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
```diff
@@ -0,0 +1,374 @@
...
+        expect(found).toBeDefined();
+        return found;
+      },
+      { timeout: 2000 },
+    );
+    if (!span) {
+      throw new Error('Span is undefined');
+    }
+
+    await vi.waitFor(
```
david-luna: I see a pattern here:
- triggering the request
- waiting the collector for something
- do expectations

IIUC from https://main.vitest.dev/guide/recipes/wait-for.html#vi-waitfor-wait-and-capture-the-value the expectations are not considered but only thrown errors. This code conditional throwing just after a `.toBeDefined()` expectation kind of confirms it. Then why we do not throw if the expectations below are not correct?

Also the documentation example uses `vi.waitFor` to get something the test needs in an async way (a database client) and after that it does the assertions normally in the test block. A gist would be


```js
// Helper functions
function getSpanOrThrow(url) {
  const span = collector.getSpans().find((s) => {
     return s.attributes.find((a) => a.key === ATTR_URL_FULL)?.value.stringValue === url;
  });
  if (span) return span;
  throw new Error(`Span with url "${url}" not found.`);
}

function getLogOrThrow(name, url) {
  const log = collector.getLogs().find((l) => {
    return l.eventName === name &&
      l.attributes.find((a) => a.key === ATTR_URL_FULL)?.value.stringValue === url;
  });
  if (log) return log;
  throw new Error(`Span with name "${name}" and url "${url}" not found.`);
}

// THEN in the test
it('should ...', () => {
  // Set url and ther vars
  await fetch(url);

  // collect the span
  const span = await vi.waitFor(() => getSpanOrThrow(url), { timeout: 5000 });
  // do expectations as usual
  expect(span).toBeDefined(); // <- maybe redundant since is handled in the waitFor

  const eventName = 'browser.resource_timing';
  const log =  await vi.waitFor(() => getLogOrThrow(eventName, url), { timeout: 5000 });
  
  expect(log).toBeDefined(); // <- also maybe redundant
  expect(log.traceId).toBe(span.traceId);
  expect(log.spanId).toBe(span.spanId);
});

```

david-luna (commented):
Thanks for working on this. The setup file is very convenient :)

e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
joaquin-diaz: You're right. Removed

e2e-tests/packages/instrumentation/fetch/instrumentation.test.ts
joaquin-diaz: Looks much cleaner now, I also removed the unnecessary throw and kept the `expect(log/span).toBeDefined()` assertion

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -167,8 +167,8 @@ export class FetchInstrumentation extends InstrumentationBase<FetchInstrumentati
           function endSpanOnError(span: Span, error: FetchError) {
             instrumentation._applyAttributesAfterFetch(span, options, error);
             instrumentation._endSpan(span, {
-              status: error.status || 0,
-              statusText: error.message,
+              statusText: error.name || error.message,
+              aborted: options.signal?.aborted ?? false,
```
overbalance: `AbortSignal.timeout()` also flips `signal.aborted`, so a timed out request takes the aborted branch in `_endSpan` and ends with no error status, no `error.type` and no status code, same as a success. We should probably classify from the rejection instead: `error.name === 'AbortError'` is a user cancel, `'TimeoutError'` and anything else is a failure.

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -167,8 +167,8 @@ export class FetchInstrumentation extends InstrumentationBase<FetchInstrumentati
           function endSpanOnError(span: Span, error: FetchError) {
             instrumentation._applyAttributesAfterFetch(span, options, error);
             instrumentation._endSpan(span, {
-              status: error.status || 0,
-              statusText: error.message,
+              statusText: error.name || error.message,
```
overbalance: When `error.name` is empty this falls back to the message, and messages are free text, so they can be high cardinality. Should we truncate or normalize like normalizeHttpRequestMethod?

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -207,7 +207,7 @@ export class FetchInstrumentation extends InstrumentationBase<FetchInstrumentati
                           'Failed to finalize span after body read',
                           e,
                         );
-                        instrumentation._endSpan(span, { status: 0 });
+                        instrumentation._endSpan(span, {});
```
overbalance: `response` is in scope here (and in the catch at line 229), so we know the request itself succeeded. Passing `{}` makes `_endSpan` treat it as "no response received" and export a 200 as an ERROR span with no status code. We should probably pass `{ status: response.status }` in both places.

packages/instrumentation/src/fetch/instrumentation.ts
```diff
@@ -207,7 +207,7 @@ export class FetchInstrumentation extends InstrumentationBase<FetchInstrumentati
...
                       }
                     },
                     (error) => {
```
overbalance, line 216: I know it's not your code but the rejection path is the one place where a response was already received before the failure, but `endSpanOnError` only forwards `statusText`, so the span loses `http.response.status_code`. Semconv sets the status code whenever a response arrived. Would it be worth passing `response.status` through here?

packages/instrumentation/src/fetch/instrumentation.ts
david-luna: This happened in `undici` instrumentation as well. This was the proposed change to reduce the cardinality https://github.com/open-telemetry/opentelemetry-js-contrib/pull/3659/changes

joaquin-diaz: To be honest based on the semconv
> [5] error.type: If the request fails with an error before response status code was sent or received, error.type SHOULD be set to exception type (its fully-qualified class name, if applicable) or a component-specific low cardinality error identifier.

> If response status code was sent or received and status indicates an error according to [HTTP span status definition](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md), error.type SHOULD be set to the status code number (represented as a string), an exception type (if thrown) or a component-specific error identifier.

> The error.type value SHOULD be predictable and SHOULD have low cardinality. Instrumentations SHOULD document the list of errors they report.

> The cardinality of error.type within one instrumentation library SHOULD be low, but telemetry consumers that aggregate data from multiple instrumentation libraries and applications should be prepared for error.type to have high cardinality at query time, when no additional filters are applied.

If the request has completed successfully, instrumentations SHOULD NOT set error.type.

I'm leaning towards not defaulting to error.message, and right now we prefer setting String(response.status):
```
const errorType =
  response.status !== undefined
    ? String(response.status)
    : response.statusText;
span.setAttribute(ATTR_ERROR_TYPE, errorType);
```

Old code doesn't even try to get error.name / error.message:

```
span.setAttribute(ATTR_ERROR_TYPE, String(response.status));
```

packages/instrumentation/src/fetch/instrumentation.ts
joaquin-diaz: Yeah, worth passing it down everywhere we have it. Added it

david-luna (approved):

overbalance (approved):

david-luna:
Thanks @joaquin-diaz! I'm going to merge since these changes are going to help me with the xhr instrumentation 😀

PR #393, opened 2026-08-12
feat(sdk): wire default context manager and propagators into startTracesSdk

### LINKED ISSUE #390: Wire default context manager and propagators into browser SDK startup
The SDK defines browser defaults for context management and propagation (getDefaultContextManager(), getDefaultPropagators()), but `startTracesSdk` currently only installs them when the user passes explicit config.

As a result, the default SDK startup does not install a global context manager or global propagator.

We should wire these defaults into SDK startup so the browser SDK behaves as the high-level SDK entry point it is intended to be.

Proposed behavior:

- If config.contextManager is provided, install that.
- Otherwise install the default browser context manager.
- If config.propagators is provided, install a CompositePropagator using those propagators.
- Otherwise install a CompositePropagator using getDefaultPropagators().
- Call contextManager.enable() before registering it.
- Add tests covering default registration and user overrides.

Note: `getDefaultSampler()` should probably not be wired as-is. The underlying TracerProvider already defaults to `ParentBasedSampler({ root: AlwaysOnSampler })`, while the helper currently returns plain `AlwaysOnSampler`. Wiring that helper would change sampling behavior. This issue should either leave sampler behavior unchanged or update/remove the helper separately.

hugolevyy:
Fixes #390

## Summary

`startTracesSdk` was only installing a context manager and propagators when the user explicitly passed them in config. Calling `startTracesSdk({})` with default config left no global context manager or propagator registered, silently breaking trace context propagation and W3C header injection.

This PR wires the existing defaults into SDK startup by following the proposed behavior (quoting the issue):

- If config.contextManager is provided, install that.
- Otherwise install the default browser context manager.
- If config.propagators is provided, install a CompositePropagator using those propagators.
- Otherwise install a CompositePropagator using getDefaultPropagators().
- Call contextManager.enable() before registering it.
- Add tests covering default registration and user overrides:
    - Added a test verifying default registration when no config is provided
    - Added a test verifying user-provided context manager and propagators take precedence
    - `afterEach` now cleans up `context` and `propagation` global state to prevent test cross-contamination

### DIFF
--- a/packages/sdk/src/traces/startTracesSdk.test.ts
+++ b/packages/sdk/src/traces/startTracesSdk.test.ts
@@ -3,7 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { diag, trace } from '@opentelemetry/api';
+import type { ContextManager, TextMapPropagator } from '@opentelemetry/api';
+import {
+  context,
+  diag,
+  propagation,
+  ROOT_CONTEXT,
+  trace,
+} from '@opentelemetry/api';
 import {
   AlwaysOffSampler,
   SimpleSpanProcessor,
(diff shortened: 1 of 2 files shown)


david-luna (approved):
thanks for you contribution :)

PR #396, opened 2026-08-14
feat: add element timing instrumentation

joaquin-diaz:
## Summary
- Add `ElementTimingInstrumentation`, observing the Element Timing API and emitting a log record per entry (identifier, element tag, render/load/start time, url, natural width/height), following the same `InstrumentationBase` + log-emission pattern as `resource-timing`/`navigation-timing`.
- Add a reusable `createPerformanceObserver`/`isEntryTypeSupported` utility (`src/utils/performanceObserver.ts`) for building single-entry-type `PerformanceObserver`s, and use it in the new instrumentation.


joaquin-diaz:
Corresponding semantic convention PR: open-telemetry/semantic-conventions#4011

PR #402, opened 2026-08-19
feat: add e2e migration parity tests for fetch instrumentation

joaquin-diaz:
## Summary
- Adds an e2e suite (`e2e-tests/migration/fetch-migration/`) that validates the fetch instrumentation migration from `@opentelemetry/instrumentation-fetch` (opentelemetry-js) to this repo's own `packages/instrumentation/src/fetch` didn't regress span shape.
- Each scenario (success, non-2xx response, no-body/204, network error, intentional abort, timeout) runs the same fetch call against a real, unmodified copy of the old instrumentation and the new one — each bootstrapped in its own iframe (since both patch the global `fetch` and can't share a realm) — and compares the resulting spans exported to the same mock OTLP collector.
- Comparison is against an explicit allowlist of semconv fields, not byte-identical spans: reading both implementations' source turned up that the old package never classifies network errors, aborts, or timeouts as span errors (`_endSpan` only marks ERROR when a synthetic `status >= 400`, which exception paths never satisfy), while the new implementation fixes this. Those three scenarios assert the specific old vs. new values directly, with comments citing where the divergence comes from, rather than asserting false equivalence.

## Test plan
- [x] `npm run test:e2e` passes locally (22/22 tests across all e2e suites, including the new one)
- [x] Pre-commit hooks (build, check-types, eslint, commit-msg) pass


e2e-tests/migration/fetch-migration/compare-spans.ts
```diff
@@ -0,0 +1,60 @@
...
+  ATTR_SERVER_PORT,
+} from '@opentelemetry/semantic-conventions';
+import { expect } from 'vitest';
+import type { OtlpKeyValue, OtlpSpan } from '../../utils/test-collector.ts';
+
+export function attrOf(span: OtlpSpan, key: string): OtlpKeyValue['value'] {
+  return span.attributes.find((a) => a.key === key)?.value ?? {};
+}
+
+const CORE_REQUEST_ATTRS = [
```
david-luna, line 21: [question] should we add here `ATTR_FULL_URL`?

e2e-tests/migration/fetch-migration/old.entry.ts
```diff
@@ -0,0 +1,39 @@
+/*
```
david-luna, line 1: I wonder if possible to reuse here `testSdkSetup` which I think it's doing the same setup regarding the traces signal (it also includes logs).  Unless we're going to break the instrumentation interface I think is okay to use the new SDK with former instrumentations.

e2e-tests/migration/fetch-migration/compare-spans.ts
joaquin-diaz: Yeah, I had to change how we fetch spans from the mock collector but now added it to the list

e2e-tests/migration/fetch-migration/old.entry.ts
joaquin-diaz: Yep, good call. Context manager is not the same but I don't think it matters for the goal of this e2e test

e2e-tests/migration/fetch-migration/fetch-migration.test.ts
```diff
@@ -0,0 +1,234 @@
...
+      '/e2e-tests/migration/fetch-migration/new.html',
+    ));
+  });
+
+  afterAll(() => {
+    oldIframe.remove();
+    newIframe.remove();
+  });
+
+  beforeEach(() => {
```
david-luna, line 125: [nit] could this be moved up to the `beforeAll` section?

david-luna (approved):
@joaquin-diaz thanks for adding this. LGTM and I've added a small question.

Notice: I'll probably move some parts to a shared folder in a follow up PR to add similar tests for XHR instrumentation.

e2e-tests/migration/fetch-migration/fetch-migration.test.ts
joaquin-diaz: No, the server is reseted after each test to avoid any conflicts between calls that we might want to assert on
```
// e2e-setup.ts

afterEach(() => {
  collector.reset();
  mockServer.resetHandlers();
});
```

joaquin-diaz:
> Notice: I'll probably move some parts to a shared folder in a follow up PR to add similar tests for XHR instrumentation.

Absolutely, go for it! Thanks for the review

Issue #403, opened 2026-08-20
Manage semantic conventions separately from instrumentation code


martinkuba:
This proposes defining the semantic conventions our instrumentations emit as a machine-readable registry in this repository, and generating TypeScript constants and documentation from it.

Today they exist only as manually-defined TypeScript constants in `semconv.ts` files in each instrumentation. Most of those attributes are not defined upstream at all (earlier proposal PRs were closed as stale).

This has a few consequences:

1. They are bare strings, not definitions (no type, requirement level, example, or record of which event an attribute belongs to).
2. Nothing validates or enforces them.
3. Code and documentation generation isn't possible.

[OTEP 4815](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4815-semantic-conventions-schema-v2.md) gives us a path forward - define these as a federated semantic convention registry in this repository, declaring the upstream semantic conventions as a dependency. OpenTelemetry Android has done this already ([opentelemetry-android#1814](https://github.com/open-telemetry/opentelemetry-android/issues/1814)).

Why not contribute them upstream instead? The semantic-conventions [CONTRIBUTING guide](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) says conventions specific to a single runtime or narrowly scoped implementation belong in the corresponding repository, and most of ours are browser-only. Federating here also lets us move without requiring broad consensus for every change, and gives us a concrete starting point for the client-side registry proposed in [community#3594](https://github.com/open-telemetry/community/issues/3594).

## Requirements

1. Semantic conventions are defined in YAML in this repository, declaring the upstream OpenTelemetry semantic conventions as an exact-version dependency.
2. The TypeScript attribute and event-name constants the instrumentations use are generated from that registry with the exception of attributes available in the `@opentelemetry/semantic-conventions` package.
3. CI fails if the generated output committed to the repository does not match the model it was generated from.
4. CI enforces naming rules on our definitions (weaver `check/naming_conventions`).
5. CI enforces that a signal never claims higher stability than an attribute it references (weaver `check/stability`).
6. CI catches removals and breaking renames of conventions we previously released (weaver `check/backwards-compatibility`).
7. Attributes that are defined in the core registry and are stable should be imported from the `@opentelemetry/semantic-conventions` package, not duplicated here.
8. A new upstream semantic conventions release is surfaced automatically rather than noticed by hand.
[...]

PR #404, opened 2026-08-20
chore: add semantic convention registry

martinkuba:
This is the first part of #403 

This adds the semantic conventions this repo emits as a federated weaver registry under `semconv/` folder. The model describes what the instrumentations emit today.

Nothing is generated from the registry yet. Tooling, the generated constants, and CI enforcement will follow in separate PRs.

Validated with `weaver registry check` (see `semconv/README.md` for the command).

Disclosure: The yaml files were generated using an AI agent.

### DIFF
--- a/semconv/README.md
+++ b/semconv/README.md
@@ -1,22 +1,11 @@
 # Federated Browser Semantic Conventions
 
-This directory holds the semantic conventions that the packages in this repository emit, modelled
-as a federated semantic convention registry ([OTEP 4815](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4815-semantic-conventions-schema-v2.md)).
-The registry manifest declares the upstream
-[OpenTelemetry semantic conventions](https://github.com/open-telemetry/semantic-conventions) as an
-exact-version dependency, so our events can reference upstream attributes without redefining them.
-
-The same approach is used by
-[opentelemetry-android](https://github.com/open-telemetry/opentelemetry-android/tree/main/semconv).
+This directory holds the semantic conventions that the packages in this repository emit, modelled as a federated semantic convention registry. The registry manifest declares the upstream [OpenTelemetry semantic conventions](https://github.com/open-telemetry/semantic-conventions) as an exact-version dependency, so our events can reference upstream attributes without redefining them.
 
 ## Why these live here
 
-The semantic-conventions
-[CONTRIBUTING guide](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md)
-places conventions specific to a single runtime or a narrowly scoped implementation in the
-corresponding repository, and most of the conventions below are browser-only. Defining them here
-also gives a concrete starting point for the client-side registry proposed in
-[community#3594](https://github.com/open-telemetry/community/issues/3594).
+The semantic-conventions [CONTRIBUTING guide](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) places conventions specific to a single runtime or a narrowly scoped implementation in the
+corresponding repository, and most of the conventions below are browser-only.
 
 Names already covered upstream are deliberately not redefined. The `browser.web_vital` and
 `exception` events are imported from the dependency registry by name, and the `error.*`, `http.*`,
@@ -36,25 +25,11 @@ model/
 
 ## Checking the registry
 
-Weaver is not required to install, build, or check the packages in this repository. To validate the
-model, run the [`weaver`](https://github.com/open-telemetry/weaver) CLI over `model/`:
+Weaver is not required to install, build, or check the packages in this repository. To validate the model, run the [`weaver`](https://github.com/open-telemetry/weaver) CLI over `model/`:
 
 ```bash
 docker run --rm -v "$PWD/semconv/model:/home/weaver/model" \
   otel/weaver:v0.25.1 registry check -r /home/weaver/model
 ```
 
-This resolves every `ref` against the pinned upstream version, so it catches unknown attribute
-references, malformed enum members, and duplicate ids.
-
-Do not pass `--future`: at the pinned upstream version it promotes the upstream registry's own
-`definition/2` file-format warnings to errors, which says nothing about this model.
-
-## Status
-
-The model currently describes what the instrumentations emit today, with no new conventions and no
-renames. Generating the TypeScript attribute and event-name constants from it, and enforcing the
-model in CI, are tracked in
-[#403](https://github.com/open-telemetry/opentelemetry-browser/issues/403). Until then the
-constants in the `semconv.ts` files under `packages/` remain hand-written, and changing an
-attribute means changing both.
+This resolves every `ref` against the pinned upstream version, so it catches unknown attribute references, malformed enum members, and duplicate ids.
--- a/semconv/README.md
+++ b/semconv/README.md
@@ -8,8 +8,15 @@ The semantic-conventions [CONTRIBUTING guide](https://github.com/open-telemetry/
 corresponding repository, and most of the conventions below are browser-only.
 
 Names already covered upstream are deliberately not redefined. The `browser.web_vital` and
-`exception` events are imported from the dependency registry by name, and the `error.*`, `http.*`,
-`server.*`, `session.*`, and `url.*` attributes we emit are referenced from it.
+`exception` events are imported from the dependency registry by name, and the `http.request.body.size`,
+`session.id`, and `url.full` attributes we emit are referenced from it.
+
+Upstream conventions that are already stable are not modelled here at all. Their constants ship in
+the root export of [`@opentelemetry/semantic-conventions`](https://www.npmjs.com/package/@opentelemetry/semantic-conventions),
+so the instrumentations import them directly and this registry has nothing to add. What it does
+cover is the attributes we emit that upstream has not stabilised, which the package exports only
+from its `/incubating` entry point — today `http.request.body.size` on the fetch and XHR spans,
+and `session.id` on everything the SDK emits.
 
 ## Layout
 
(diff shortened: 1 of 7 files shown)


semconv/model/browser/events.yaml
```diff
@@ -0,0 +1,175 @@
...
+      - ref: browser.resource_timing.response_end
+        requirement_level: required
+      - ref: browser.resource_timing.transfer_size
+        requirement_level: required
+      - ref: browser.resource_timing.encoded_body_size
+        requirement_level: required
+      - ref: browser.resource_timing.decoded_body_size
+        requirement_level: required
+      - ref: browser.resource_timing.next_hop_protocol
+        requirement_level: recommended
```
overbalance, line 153: `_emitResource` sets this one unconditionally, same as everything above it. On a cross-origin resource with no `Timing-Allow-Origin` it arrives as `""` rather than missing, which the note on the attribute already covers. So I think it is `required` in practice.

`render_blocking_status` below is a real `recommended`, it is `undefined` outside Chromium, which is what the `@ts-expect-error` next to it is about.

```suggestion
      - ref: browser.resource_timing.next_hop_protocol
        requirement_level: required
```

semconv/model/browser/events.yaml
```diff
@@ -0,0 +1,175 @@
...
+        requirement_level: required
+
+  - id: event.browser.resource_timing
+    type: event
+    name: browser.resource_timing
+    stability: development
+    brief: >
+      This event represents the loading of a single resource by the page, as reported by the
+      [PerformanceResourceTiming](https://developer.mozilla.org/docs/Web/API/PerformanceResourceTiming)
+      API.
```
overbalance, line 113: Two things about this group only live in a `#` comment in `registry.yaml` (lines 195-199), and weaver does not render those.

The units. Each `browser.navigation_timing.*` brief says "Time, in milliseconds relative to the start of the navigation" inline, so it survives into the generated docs. The `browser.resource_timing.*` briefs just say "Time immediately before the browser starts to fetch the resource", so this event ends up documented with no unit and no reference point. I think it is also the time origin of the document rather than the navigation, which only differs in a worker but is the more accurate wording.

The other is `Timing-Allow-Origin`. Only `next_hop_protocol` mentions it today, but without that header the browser zeroes `connect_start`, `connect_end`, `request_start`, `response_start`, the redirect timings and all three sizes, and `_emitResource` copies them straight through. Someone reading `connect_end: 0` will read it as an instant connection.

Could we put both in a group note here? `weaver registry check` passes with this.

```suggestion
    brief: >
      This event represents the loading of a single resource by the page, as reported by the
      [PerformanceResourceTiming](https://developer.mozilla.org/docs/Web/API/PerformanceResourceTiming)
      API.
    note: >
      The timings are in milliseconds, relative to the time origin of the document.
      For a cross-origin resource that is not served with a `Timing-Allow-Origin` response header
      field, the browser reports zero for most of these timings and sizes, so a zero here can mean
      the value is not available rather than that the step took no time.
```

semconv/model/browser/registry.yaml
```diff
@@ -0,0 +1,422 @@
...
+          Soft navigations performed by a single-page application through the History or Navigation
+          API are same-document navigations; the initial page load is not.
+        examples: [true, false]
+
+      - id: browser.navigation.hash_change
+        type: boolean
+        stability: development
+        brief: >
+          Whether the navigation only changed the fragment of the URL.
+        examples: [true, false]
```
overbalance, line 57: `isHashChange` ends on `((fromHasHash && toHasHash) || (!fromHasHash && toHasHash))`, which reduces to just `toHasHash`, and the jsdoc above it says that is on purpose (added or changed, not removed). So `/page#a` -> `/page` changes only the fragment and still reports `false`.

The code looks deliberate, so we should probably match it here instead. This brief is the thing people will write queries against.

```suggestion
        brief: >
          Whether the navigation added or changed the fragment of the URL.
        note: >
          A navigation that removes the fragment is not reported as a hash change, even though
          only the fragment changed.
        examples: [true, false]
```

semconv/model/browser/registry.yaml
```diff
@@ -0,0 +1,422 @@
...
+              brief: The auxiliary button, usually the mouse wheel button.
+            - id: right
+              value: "right"
+              stability: development
+              brief: The secondary button, usually the right button.
+        stability: development
+        brief: >
+          The mouse button that was pressed, as reported by
+          [`MouseEvent.button`](https://developer.mozilla.org/docs/Web/API/MouseEvent/button).
+        examples: ["left"]
```
overbalance, line 382: `MouseEvent.button` is a number, and `_getMouseButtonFromMouseEvent` folds 3 (back) and 4 (forward) into `left` through the `default` branch. So a back click is recorded as a left click, and "as reported by" reads like we pass the DOM value through.

Adding `back` and `forward` members would be the other way to fix it, but that changes what we emit, so I think it belongs in a follow up rather than here.

```suggestion
        brief: >
          The mouse button that was pressed, derived from
          [`MouseEvent.button`](https://developer.mozilla.org/docs/Web/API/MouseEvent/button).
        note: >
          Only the primary, auxiliary and secondary buttons have their own value. The browser back
          and forward buttons are reported as `left`.
        examples: ["left"]
```

semconv/README.md
```diff
@@ -0,0 +1,42 @@
+# Federated Browser Semantic Conventions
+
+This directory holds the semantic conventions that the packages in this repository emit, modelled as a federated semantic convention registry. The registry manifest declares the upstream [OpenTelemetry semantic conventions](https://github.com/open-telemetry/semantic-conventions) as an exact-version dependency, so our events can reference upstream attributes without redefining them.
+
+## Why these live here
+
+The semantic-conventions [CONTRIBUTING guide](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) places conventions specific to a single runtime or a narrowly scoped implementation in the
+corresponding repository, and most of the conventions below are browser-only.
```
overbalance, line 8: Nothing follows this, the next thing is `## Layout`. Did a list get dropped, or should it point at the registry instead?

The guide also says "runtime, library, or narrowly scoped implementation" and phrases it as a SHOULD, so this is a little stronger than the original.

```suggestion
The semantic-conventions [CONTRIBUTING guide](https://github.com/open-telemetry/semantic-conventions/blob/main/CONTRIBUTING.md) recommends that conventions specific to a single runtime, library, or a narrowly scoped implementation live in the
corresponding repository, and most of the conventions in this registry are browser-only.
```

semconv/README.md
```diff
@@ -0,0 +1,42 @@
...
+Names already covered upstream are deliberately not redefined. The `browser.web_vital` and
+`exception` events are imported from the dependency registry by name, and the `http.request.body.size`,
+`session.id`, and `url.full` attributes we emit are referenced from it.
+
+Upstream conventions that are already stable are not modelled here at all. Their constants ship in
+the root export of [`@opentelemetry/semantic-conventions`](https://www.npmjs.com/package/@opentelemetry/semantic-conventions),
+so the instrumentations import them directly and this registry has nothing to add. What it does
+cover is the attributes we emit that upstream has not stabilised, which the package exports only
+from its `/incubating` entry point — today `http.request.body.size` on the fetch and XHR spans,
+and `session.id` on everything the SDK emits.
```
overbalance, line 19: "What it does cover is" followed by two attributes reads like that is the whole registry, but `registry.yaml` defines about 40 `browser.*` attributes that upstream has no definition for at all, and those are not in `/incubating` either. The two named here are the upstream non-stable ones we `ref`.

Could we split it into two sentences, one for the attributes we define because upstream has none, one for the upstream non-stable ones we reference?

semconv/model/browser/common.yaml
```diff
@@ -0,0 +1,12 @@
+groups:
+  - id: attributes.browser.session
```
overbalance, line 2: Nothing refs `attributes.browser.session`, so `session.id` will not appear in the generated docs for any of the events it is actually on. Is that intentional for now, with the event groups picking it up in the generation PR, or should the groups ref it here?

Smaller one on the brief below: it says the SDK adds these to the telemetry it emits, but `SessionSpanProcessor` and `SessionLogRecordProcessor` are opt-in factories and `startBrowserSdk` does not wire them up. The note's "When session tracking is enabled" covers it, so maybe the brief can say the same.

semconv/model/manifest.yaml
```diff
@@ -0,0 +1,7 @@
+name: opentelemetry-browser
+description: Local semantic conventions for opentelemetry-browser.
+schema_url: https://opentelemetry.io/schemas/opentelemetry-browser/0.1.0
+dependencies:
+  - name: otel
+    registry_path: https://github.com/open-telemetry/semantic-conventions@v1.44.0[model]
```
overbalance, line 6: Worth a line saying why this is v1.44.0 while the packages are on `^1.43.0`? It has to be 1.44.0, because `events.yaml` imports `browser.web_vital` and the installed 1.43.0 ships `EVENT_BROWSER_WEB_VITAL` but none of the `browser.web_vital.*` attributes (I grepped the build output). Without a note it reads like drift, and someone may pin it back to match the dependency later.

semconv/README.md
```diff
@@ -0,0 +1,42 @@
+# Federated Browser Semantic Conventions
+
+This directory holds the semantic conventions that the packages in this repository emit, modelled as a federated semantic convention registry. The registry manifest declares the upstream [OpenTelemetry semantic conventions](https://github.com/open-telemetry/semantic-conventions) as an exact-version dependency, so our events can reference upstream attributes without redefining them.
```
overbalance, line 3: Should we standardize on American spelling for these, to match the rest of the docs? `modelled` is here and on line 14, and `stabilised` is on line 17.

```suggestion
This directory holds the semantic conventions that the packages in this repository emit, modeled as a federated semantic convention registry. The registry manifest declares the upstream [OpenTelemetry semantic conventions](https://github.com/open-telemetry/semantic-conventions) as an exact-version dependency, so our events can reference upstream attributes without redefining them.
```

semconv/model/browser/registry.yaml
martinkuba: I have created this issue to address it in the instrumentation
https://github.com/open-telemetry/opentelemetry-browser/issues/438

martinkuba:
Converting to draft, while we are waiting for a decision whether these should live in this repository or in https://github.com/open-telemetry/semantic-conventions-client-side.

PR #421, opened 2026-09-07
feat(sdk): populate browser.document.url.full on all spans and log records

mquentin:

https://github.com/open-telemetry/opentelemetry-browser/issues/174.

A browser app needs to know *which page* a piece of telemetry came from.

`browser.document.url.full` semantic has been approved already https://github.com/open-telemetry/semantic-conventions/pull/3633

Like session.id this field `browser.document.url.full` is meant to be carried by all spans and logs which led me to implement a dedicated span and log processor.

**An entity effort has been in discussion for a while for the SIG but my take is that having these new `document` processors Demoed in the sandbox on top of the `session.id` will help making a case and will help shape the entity definition**

Adds a `document` module to `@opentelemetry/browser-sdk`, exposed as the
`@opentelemetry/browser-sdk/document` subpath, following the same pull model as
`session/`:

- `DocumentProvider` — `getDocumentUrl(): string | null`
- `DocumentSpanProcessor` / `DocumentLogRecordProcessor` — set the attribute in
  `onStart` / `onEmit`
- `LocationDocumentProvider` — reads `location.href`, with an optional
  `sanitizeUrl` hook to trim query parameters and keep cardinality down

Because the attribute is applied by processors rather than by an instrumentation,
coverage is universal — every instrumentation plus user code, with no opt-in.

The sandbox (`sandbox/src/otel.ts`) registers both processors alongside the existing
session ones.

As a result here is a screenshot of the processor demoed in the sandbox:

Screenshot 2026-09-07 at 16 37 25

Note on timing: spans capture the URL at `onStart` and log records at `onEmit`, so a
span outliving a soft navigation keeps the URL of the page it started on. Flagged
with a TODO in case we later want the end-of-span URL as well.


martinkuba (approved):
This looks good to me overall. We might eventually switch to an actual entity/resource attribute, but the SDK cannot handle that yet. The package is experimental, so it changing exports is fine later, I think.

PR #425, opened 2026-09-09
docs(sdk): correct config option docs

overbalance:

Several JSDoc comments on the SDK config types describe the opposite of what the code does. `LogsConfig.processors` and `TracesConfig.processors` say that setting them makes the SDK ignore `exportConfig`, but `startLogsSdk` and `startTracesSdk` still create a batch processor when the signal sets its own `exportConfig`. `TracesConfig.batchProcessorConfig` claims to enable a `BatchSpanProcessor` on its own, which it does not.

These comments are what make the propagation rules easy to get wrong, so they are worth correcting on their own.

Comments only. No behavior changes, and no test changes other than one comment.

The `processors`, `exportConfig` and `batchProcessorConfig` docs on `RootConfig`, `LogsConfig` and `TracesConfig` now match the signal SDKs. `ExportConfig.url` says that a URL set on a signal is used as is, while a signal `exportConfig` with no `url` gets the root URL with the signal path set on it. `QuickStartConfig.exportUrl` says the signal path replaces any path given.

The `combineSdks` doc had an unfinished sentence. A test note said the suite mocks provider registration, which it does not, and it now describes what the hook actually does.

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- [x] `npm test` passes repo wide, `packages/sdk` at 64/64, unchanged from `main` since no behavior changed.
- [x] `npm run check` is clean.
- [x] Each corrected comment was read against the code it describes in `combineSdks`, `startLogsSdk` and `startTracesSdk`.

## Checklist:

### DIFF
--- a/packages/sdk/src/core/sdk.ts
+++ b/packages/sdk/src/core/sdk.ts
@@ -21,7 +21,7 @@ interface SdkFactories {
 }
 
 /**
- * Utility functions to extract the configurations from the factory
+ * Utility types to extract the configurations from the factory
  * functions and remove the common properties (which will be already
  * available at the config root)
  */
@@ -40,8 +40,8 @@ const DEFAULT_CONFIG: RootConfig = {
 const NOOP_SDK = { shutdown: () => Promise.resolve() };
 
 /**
- * Combines different SDK factory functions into a single one
- * which accepts a global configuration along
+ * Combines different SDK factory functions into a single one which accepts a
+ * root configuration shared by every signal
  */
 export function combineSdks<T extends SdkFactories>(
   factories: T,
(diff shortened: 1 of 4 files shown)


packages/sdk/src/core/sdk.ts
```diff
@@ -99,9 +99,8 @@ export function combineSdks<T extends SdkFactories>(
       const logsConfig = (config?.logs || {}) as LogsConfig;
       const isGenericEndpoint = !logsConfig.exportConfig?.url;
 
-      // Propagate root configs to signal configs only when the signal does not
-      // have custom processors. When processors are provided, exportConfig and
-      // batchProcessorConfig are intentionally ignored per the LogsConfig docs.
+      // Propagate root configs only when the signal has no custom processors.
```
david-luna, line 102: I'm okay with this change but it will conflict with the one proposed in https://github.com/open-telemetry/opentelemetry-browser/pull/357

I suggest we resolve and merge that PR 1st and then apply the rest of the changes here

overbalance: Closing for now. I forgot about the config changes in flight.

PR #429, opened 2026-09-14
feat(instrumentation): add Long Animation Frames instrumentation

### LINKED ISSUE #401: Create LoAF Instrumentation
### What browser API or library should be instrumented?

[LongAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/Performance_API/Long_animation_frame_timing)

### Is this commonly used?

Yes

### Additional context

Code exists, just need porting.

SyedTashfin:
## Summary

Ports the Long Animation Frames instrumentation from `opentelemetry-js-contrib` (open-telemetry/opentelemetry-js-contrib#3751) into this repository. Closes #401.

Long Animation Frames is the successor to the deprecated Long Tasks API. This instrumentation observes `long-animation-frame` performance entries and emits each one as a log record — matching the existing `console`/`errors` pattern — carrying the frame timing breakdown (`blockingDuration`, `renderStart`, `styleAndLayoutStart`, `firstUIEventTimestamp`) and the per-script attribution (`scripts`).

## Changes

- `packages/instrumentation/src/long-animation-frame/` — instrumentation, types, semantic attribute constants, and tests.
- `packages/instrumentation/package.json` — adds the `./experimental/long-animation-frame` export and keyword.

## Notes

- Emits logs rather than spans, consistent with the console/errors instrumentations here (the version originally opened in js-contrib was span-based).
- Includes an `applyCustomLogRecordData` hook, the log-record equivalent of the js-contrib version's `observerCallback`.

## Testing

10 vitest cases covering buffered-entry emission, log timestamping, the custom hook (including error containment), unsupported-environment guards, disable/re-enable, and observer setup failures.

<details>
<summary>Generative AI disclosure</summary>

I used Hermes Agent to assist with repository research, porting the code to this repository's conventions, and test drafting. I reviewed the final change and take responsibility for it.
</details>


packages/instrumentation/src/long-animation-frame/semconv.ts
david-luna, line 1: note: to be discussed in browser SIG and added in the [registry](https://github.com/open-telemetry/opentelemetry-browser/pull/404)

packages/instrumentation/src/long-animation-frame/instrumentation.ts
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  override disable(): void {
+    this._isEnabled = false;
+    this._observer?.disconnect();
+    this._observer = undefined;
+  }
+
+  private _isSupported(): boolean {
+    if (
```
david-luna, line 89: I've seen this in other instrumentations (more than two). IMO this is a candidate to be a utility function. out of the scope of this PR

packages/instrumentation/src/long-animation-frame/instrumentation.ts
```diff
@@ -0,0 +1,182 @@
...
+                    script.forcedStyleAndLayoutDuration,
+                  window_attribution: script.windowAttribution,
+                }),
+              ),
+            }
+          : {}),
+      },
+    };
+
+    const hook = this.getConfig().applyCustomLogRecordData;
```
david-luna, line 156: If the apply function does get additional context is easier to use a `LogRecordProcessor` and do not more complexity to the instrumentation. ref: https://github.com/open-telemetry/opentelemetry-browser/issues/359#issuecomment-5282391676

overbalance:
@SyedTashfin The ticket is assigned to me because the instrumentation code is already completed. I think the cardinality of the code in your PR may be a bit high. I'll review in depth later.

packages/instrumentation/src/long-animation-frame/instrumentation.ts
david-luna: @SyedTashfin 

I'd suggest to remove it. The entry contains all the context and is translated to a log. There is no user action attached or event related to it. This can be done with a `LogRecordProcessor` when configuring the `LoggerProvider`

david-luna (changes requested):

packages/instrumentation/src/long-animation-frame/semconv.ts
martinkuba: I would not hold this PR on this. We can add these to the registry once it is ready.

SyedTashfin:
Removed the `applyCustomLogRecordData` hook in 13bee56. Agreed that a `LogRecordProcessor` on the `LoggerProvider` does the same job, and the record already carries the whole entry, so the hook was redundant.

Worth flagging: `navigation`, `user-action` and `web-vitals` in this repo still expose the same hook. I took it out here as asked, but if you'd rather keep the instrumentations consistent I can put it back.

I've left the semconv attribute names local to the instrumentation for now and will move them to the registry once #404 is in.

@overbalance you mentioned the ticket is yours because the instrumentation is already completed. I couldn't find a branch or a PR for it, so I can't tell what state it is in. If you're porting it, tell me and I'll close this one; I just don't want two implementations of the same thing running in parallel. If it's the size of the diff that's the concern, tell me which parts you'd cut and I'll strip it down.

packages/instrumentation/src/long-animation-frame/instrumentation.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  override enable(): void {
+    if (this._observer || !this._isSupported()) {
+      return;
+    }
+
+    const observer = new PerformanceObserver((list) => {
+      for (const entry of list.getEntries()) {
+        this._emitLongAnimationFrame(
```
mquentin: Suggestion:

could we have a `debounce` or `throttle` mechanism here if somehow too many long animation framed occurred at the same time ?

I know there are supposed to have a meaningful duration for each frames and you're not supposed to have thousands of very small frames but just in case a `throttle` might be a good safeguard.

mquentin (commented):
Thanks for your contribution, I juste have a small suggestion and I would love to have your input on a question:

## Suggestion:

I think having a small `throttle` safeguard for this instrumentation would be great but this instrumentation is indeed a good first step toward having a better performance monitoring.

## My main question is

How could we aggregate these long animation frame ? I feel most of the value of these events comes from being able to identify which user interaction generated this long animation frame and on which page did it occurs.

Do you have any take on this ?

SyedTashfin:
Answering the question first, then the throttle, because I disagree with that one.

On aggregation: I think the record is already an attribution object, so this is mostly a grouping problem rather than a data problem.

For which interaction — `firstUIEventTimestamp` is the timestamp of the first UI event processed in that frame, and a click's `timeStamp` is a `DOMHighResTimeStamp` on the same clock, so a frame can be joined to the interaction that triggered it without any context plumbing. The `scripts` array then names the culprit: `invoker: "DOMWindow.onclick"`, `invokerType: "event-listener"`, plus `sourceURL`, `sourceFunctionName`, `sourceCharPosition`, and a per-script split of `blockingDuration` against `forcedStyleAndLayoutDuration`.

The other half of that join is already in this repo: the web-vitals instrumentation timestamps the INP record with `attribution.interactionTime`, and puts the raw attribution in the body when `includeRawAttribution` is on. So matching a LoAF's `firstUIEventTimestamp` against a `browser.web_vital` INP record gives you interaction target and type together with the scripts that blocked the frame. That's the aggregation I'd build: INP tells you which interaction was slow, LoAF tells you which script made it slow.

For which page — nothing to add here. The SDK's `createDocumentLogRecordProcessor` already stamps `browser.document.url.full` on every log record, and `LocationDocumentProvider` reads `location.href` at emit time so `pushState` navigation is reflected. One caveat: with `buffered: true` the replayed entries get the current URL rather than the URL they happened on, so those first records can be mislabelled on a page that navigates early.

Where I'd keep the line is not in the instrumentation. A rollup here needs a time window and a flush policy, which turns a faithful record emitter into a metric reporter with config surface. `source_url` and `source_function_name` are unbounded — fine as log attributes, a trap as metric dimensions. If we want counters or histograms, the bounded attribute set belongs in the registry (#404) as an explicit event or metric definition, not in this PR.

On the throttle: I'd rather not, and I think it would actively hurt. LoAF is a jank signal — entries only exist for frames past 50ms, and the browser caps the timeline buffer at 200 entries. A throttle drops the worst frames first: a two second jank that produces dozens of entries would report one, and the p95 downstream gets better while the page gets worse. That's the one failure mode I don't want from a performance instrumentation.

If you want a hard ceiling, I'd do it with accounting instead: a per-callback cap, plus a dropped count on the next record so the loss is visible. Or reuse the shape `resource-timing` already has — `batchSize`, `maxQueueSize`, `forceProcessingAfter`, drained on an idle callback. Either is a follow-up; a first version doesn't need it.

